PR #11, opened 2019-08-27
Message ID and headers

SeanFeldman:
This PR is doing the following:
- Ensure we have a valid message ID from the original message before creating one
- Ensure headers are unmodified when invoking the error pipeline


src/NServiceBus.AzureFunctions.AzureStorageQueues/AzureFunctionEndpoint.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace NServiceBus.AzureFunctions.AzureServiceBus
```
timbussmann: what's with this class?

timbussmann:
rebased this on top of master

src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
```diff
@@ -62,6 +55,17 @@ public async Task Process(Message message, ExecutionContext executionContext)
 
                 throw;
             }
+
+            MessageContext CreateMessageContext(Message originalMessage)
```
timbussmann: personally I don't see much value in this method given it's only called from a single place and it doesn't do anything complex but no strong feelings about this.

src/NServiceBus.AzureFunctions.AzureStorageQueues/AzureFunctionEndpoint.cs
timbussmann: I removed it.

PR #15, opened 2019-08-28
Derive endpoint name from function name

timbussmann:
makes the API easier to use because users won't have to define a endpoint name constant or something like that but instead we can automatically derive the endpoint name from the function name.

This would also simplify the sample a litte bit further.


timbussmann:
no it can't be overridden. I was thinking about this but couldn't come up with a reason why you would do so. Using a different name would mostly break a whole lot of functionality and I think for this experimental state we can just use the function name and see whether somebody raises a use case which requires control over the endpoint name.

SeanFeldman:
Then it's not good. A Function name "My_Function" will break it for ASQ.

timbussmann:
> Then it's not good. A Function name "My_Function" will break it for ASQ.

then we should prevent users from using function names with invalid characters given that they won't be able to use local sends, delayed retries and so on?

PR #16, opened 2019-08-28
Support environment variables

timbussmann:
Note: this PR is built in top of #15 which needs to merged before.

The current configuration requires the connection string to be stored in a local.settings.json. Therefore users can't use connection strings configured in the portal as environment variables. As the local.settings.json "values" will be available as environment variables automatically, we can just fetch the name from the environment variables which makes this more flexible and useful.

I've also made the connectionstring name an optional parameter configured to the default trigger specific connection strings documented [here for ASB](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-service-bus#trigger---configuration) and [here for ASQ](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-queue#trigger---configuration).


timbussmann:
updated the PR as discussed @SeanFeldman

PR #19, opened 2019-08-29
Set recoverability settings back to the NServiceBus defaults

SeanFeldman:
Azure Functions can do immediate and delayed retries. While there's some issue with the ASQ delayed retries, ASB worked flawlessly.
Suggested changes:
- Set ASB recoverability to the defaults (5 immediate retries and 3 delayed).
- Set ASQ recoverability to what the Functions SDK can handle (4 immediate retries and 3 delayed).
  - The ASQ trigger can only handle 5 deliveries before it dead-letters the message. With the default 5 immediate retries and one normal processing that is one attempt too much.

I didn't know if there's a way to reset recoverability attempt to the defaults.
I've also contemplated removing the code that disables recoverability in NServiceBus.Serverless, but wasn't sure if that will affect AWS Lambda negatively or not. @boblangley or @danielmarbach could you please confirm if removing the code in https://github.com/ParticularLabs/NServiceBus.Serverless/blob/9664da1627f2187ef8d46feb58d4f50bd648f3ff/src/NServiceBus.Serverless/ServerlessEndpointConfiguration.cs#L23 will affect Lambda or not? Thank you.

### DIFF
--- a/src/NServiceBus.AzureFunctions.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
+++ b/src/NServiceBus.AzureFunctions.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
@@ -22,8 +22,9 @@
             var connectionString = System.Environment.GetEnvironmentVariable(connectionStringName);
             Transport.ConnectionString(connectionString);
 
-            this.AdvancedConfiguration.Recoverability().Immediate(settings => settings.NumberOfRetries(5));
-            this.AdvancedConfiguration.Recoverability().Delayed(settings => settings.NumberOfRetries(3));
+            var recoverability = AdvancedConfiguration.Recoverability();
+            recoverability.Immediate(settings => settings.NumberOfRetries(5));
+            recoverability.Delayed(settings => settings.NumberOfRetries(3));
         }
     }
 }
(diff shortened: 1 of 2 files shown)


danielmarbach:
Yes it would affect Lambda. When we did discuss the recoverability parts we moved to a model which uses the core default recoverability policy wrapped in a recoverability policy that throws if move to error is disabled. That policy also supports delayed retries. But by default the server less configuration base constructor disables delayed retries because they are not all supported natively by all the transports. For example SQS can support native delayed retries but they require FIFO queues that are currently polled by the transport. If we would want to support this we would have to enforce a second trigger per function that deals with messages in the fifo queue. 

What you did here is correct. Unfortunately there are no default constants for the number of retries if you want to reenable the core behavaior

src/NServiceBus.AzureFunctions.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
```diff
@@ -21,6 +21,9 @@ public ServiceBusTriggeredEndpointConfiguration(string endpointName, string conn
 
             var connectionString = System.Environment.GetEnvironmentVariable(connectionStringName);
             Transport.ConnectionString(connectionString);
+
+            this.AdvancedConfiguration.Recoverability().Immediate(settings => settings.NumberOfRetries(5));
```
danielmarbach: we used the following style in SQS trigger

```
var advanced = AdvancedConfiguration;
advanced.
```

or here I'd use

```
var recoverability = AdvancedConfiguration.Recoverability();
recoverability.
```

SeanFeldman:
Thank you for the confirmation @danielmarbach. I suspected that tweaking would need to be on the specific serveless offer level, but wasn't 100% sure.

PR #20, opened 2019-08-29
Disable delayed retries for ASQ

timbussmann:
Disable ASQ delayed retries for now as they don't seem to work out of the box.


SeanFeldman:
Let's hold off on merging until no have my investigation completed (cc @davidboike)

PR #21, opened 2019-08-29
Get the functions logger into the pipeline

DavidBoike:
It's not pretty, and fights a bit with the current abstraction, but it appears to work.


timbussmann:
does that work when multiple functions are invoked in parallel? Azure shows logs on a invocation based level, wouldn't this caching mess this up?

danielmarbach:
Potential alternative https://github.com/ParticularLabs/NServiceBus.AzureFunctions/pull/25

SeanFeldman:
PR #25 based on this PR is merged.

PR #23, opened 2019-08-30
Auto configuration via reflection

timbussmann:
We can use reflection to walk up the stack trace and find the trigger specific attributes. From these attributes we can use the endpoint name and optionally the connection name so users don't have to repeat this information unnecessarily.

Because the endpoint name is a ctor param on the base configuration class, I've just went with a static factory which goes on a strack trace hike so users can always fall back to passing the params manually. Not sure about the naming but I didn't come up with something better quickly.

If we go with David's PR using the shared assembly, we might be able to move some of this code to the shared assembly.

### DIFF
--- a/src/Shared/AttributeDiscoverer.cs
+++ b/src/Shared/AttributeDiscoverer.cs
@@ -28,8 +28,6 @@ namespace NServiceBus.AzureFunctions
                         }
                     }
                 }
-
-                return null;
             }
 
             return null;
--- a/src/Shared/AttributeDiscoverer.cs
+++ b/src/Shared/AttributeDiscoverer.cs
@@ -1,0 +1,38 @@
+namespace NServiceBus.AzureFunctions
+{
+    using System;
+    using System.Diagnostics;
+    using System.Reflection;
+    using Microsoft.Azure.WebJobs;
+
+    class TriggerDiscoverer
+    {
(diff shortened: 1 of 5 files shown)


src/Shared/AttributeDiscoverer.cs
```diff
@@ -0,0 +1,38 @@
+namespace NServiceBus.AzureFunctions
+{
+    using System;
+    using System.Diagnostics;
+    using System.Reflection;
+    using Microsoft.Azure.WebJobs;
+
+    class TriggerDiscoverer
```
danielmarbach: Can be static because you'll never need an instance of this class

SeanFeldman: You forgot a `nit` prefix in your comment 😝 
Absolutely correct.

danielmarbach: I thought "can" implies that

src/Shared/AttributeDiscoverer.cs
```diff
@@ -28,8 +28,6 @@ class TriggerDiscoverer
                         }
                     }
                 }
-
-                return null;
```
timbussmann: oh right, that `return null` was too late. It should have been after line 28 instead to avoid walking through the stacks further.

PR #25, opened 2019-08-30
Alternative logger

danielmarbach:
An alternative to #21 

This PR introduces `FunctionExecutionContext` that is populated with `ILogger`. Which can be provided to the configuration lambda code.


SeanFeldman:
This approach is more aligned with AzureWorks and removes the instance object which I felt could lead to some side effects. I'll make a second pass over this PR.

PR #27, opened 2019-09-02
Add basic tests

SeanFeldman:
Basic tests for NServiceBus.AzureFunctions

### DIFF
--- a/src/ServiceBus.Tests/When_message_is_received.cs
+++ b/src/ServiceBus.Tests/When_message_is_received.cs
@@ -38,31 +38,31 @@
                 return message;
             }
         }
-    }
 
-    public class TestContext
-    {
-        public int HandlerInvocationCount => count;
-
-        public void HandlerInvoked() => Interlocked.Increment(ref count);
-
-        int count;
-    }
-
-    class HappyDayMessage : IMessage {}
-
-    class HappyDayMessageHandler : IHandleMessages<HappyDayMessage>
-    {
-        TestContext testContext;
-
-        public HappyDayMessageHandler(TestContext testContext)
+        public class TestContext
         {
-            this.testContext = testContext;
+            public int HandlerInvocationCount => count;
+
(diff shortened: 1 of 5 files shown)


SeanFeldman:
TC was missing the environment variable required to read the connection strings. I've added those.

src/ServiceBus.Tests/When_message_is_received.cs
```diff
@@ -0,0 +1,68 @@
...
+                var bytes = Encoding.UTF8.GetBytes("<HappyDayMessage/>");
+                var message = new Message(bytes);
+                message.UserProperties["NServiceBus.EnclosedMessageTypes"] = typeof(HappyDayMessage).FullName;
+
+                return message;
+            }
+        }
+    }
+
+    public class TestContext
```
timbussmann: context, messages and handler classes are usually nested classes in the core acceptance tests to not pollute the namespace too much. Would that make sense too here?

SeanFeldman: I'm ok to move it, but not sure what exactly "namespace pollution" in the case of a test project means. This is not a public assembly or shared with customers.

SeanFeldman: Thought about it a bit more and here are my thoughts:

1. We can't pollute the namespace as the namespace is defined as a testing namespace. Therefore anything found in the namespace is expected to belong to the tests.
1. These are not Core ATTs. These are tests of the assembly we're not shipping.

As such, I'm good with keeping it as-is.

SeanFeldman: Discussed on the call. https://github.com/ParticularLabs/NServiceBus.AzureFunctions/pull/27/commits/e8c6eb5d174c662f84c42684320492a9556edeba

SeanFeldman:
Discussed on the sync call to add an additional test to verify retries are working as expected w/o invoking the infrastructure.

PR #31, opened 2019-09-02
Logging changes

timbussmann:
Made a few changes to the logging approach:
* Using `AsyncLocal` allows to ensure that every pipeline invocation via `Process` uses the actual logger from this invocation and avoid that a concurrent processing attempts switches the logger in the background
* The logger factory needs to be set as early as possible, therefore moving the initialization which doesn't require any "runtime" information to the configuration static ctor.

With this solution, we could also revert the wrapping context again and reduce required parameters at configuration time. That can be addressed separately though.


SeanFeldman:
> With this solution, we could also revert the wrapping context again and reduce required parameters at configuration time. 

I'd rather not. I find it cleaner to pass `FunctionExecutionContext` around rather than multiple parameters.

timbussmann:
> I'd rather not. I find it cleaner to pass FunctionExecutionContext around rather than multiple parameters.

there are no more parameters to pass around except the `ExecutionContext`

timbussmann:
updated the PR according to our sync

SeanFeldman:
Conflicts with `master`, will rebase.

PR #36, opened 2019-09-04
Verify failed messages sent to the error queue

### LINKED ISSUE #35: Add verification test recoverability is configured and works as expected for both transports
For ASB it means up to 5 immediate retries and up to 3 delayed retries by default and then the message is moved to the error queue.

For ASQ it means up to 4 immediate retries by default and then the message is moved to the error queue.

SeanFeldman:
Fixes #35

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,3 +5,4 @@ Process messages in AzureFunctions using the NServiceBus message pipeline.
 ## Running tests locally
 
 Test projects included in the solution rely on two environment variables used by Azure Functions SDK. Those are `AzureWebJobsServiceBus` and `AzureWebJobsStorage`.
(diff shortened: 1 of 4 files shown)


timbussmann:
seems like those tests fail because there are no valid connection strings. Not sure why only one of these fails those?

SeanFeldman:
This PR was discussed and the decision was:
- Document the prerequisites to run the tests locally (done in the latest commits)
- Raise an issue to review the testing approach (issue https://github.com/ParticularLabs/NServiceBus.AzureFunctions/issues/37 raised)
- Merge into `master` after 0.1.0 is released as it doesn't affect the release

PR #45, opened 2020-06-04
Add assembly scanning

timbussmann:
Adds additional assembly scanning logic to scan the functions dedicated `bin` folder which currently isn't scanned by Core's assembly scanning mechanism. All found assemblies (except known system assemblies) are then loaded into the current `AssemblyLoadContext` which is different than  `AssemblyLoadContext.Default` when running on Azure Functions.

TODO:
* [x] Add tests
* [x] Remove TODOs in code


boblangley:
@SeanFeldman and I did some troubleshooting on the failed tests. We discovered there is a problem making assumptions about the path in a test vs a published function. We added the ability for the tests to define the scan path.

This led us to tests continuing the fail because handlers are not found. The failure is reported as object is not set to an instance of an object, but if you add a breakpoint to https://github.com/Particular/NServiceBus.AzureFunctions/blob/master/src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs#L43 we discovered the cause is actually a handler not found issue. We could not figure out why this could be as we still have not setup a test to validate the reported issue (externally defined handlers). Since the core also scans, we cannot reason why the changes made here would result in this error. 

Perhaps you can continue investigating.

src/Shared/Serverless/ServerlessEndpoint.cs
```diff
@@ -95,7 +95,10 @@ void LoadAssemblies(TExecutionContext executionContext)
 
                     //TODO: skip Particular signed assemblies too?
 
-                    assemblyLoadContext.LoadFromAssemblyName(assemblyName);
+                    // LoadFromAssemblyName works when actually running inside a function as FunctionAssemblyLoadContext probes the "bin" folder for the assembly name
+                    // this doesn't work when running with a different AssemblyLoadContext (e.g. tests) and the assembly needs to be loaded by the full path instead.
+                    assemblyLoadContext.LoadFromAssemblyPath(binFile);
```
SeanFeldman, line 100: https://github.com/Particular/NServiceBus/pull/5652/files#diff-a2b31a822dbf4f05463572f427b8428dR149 😉

timbussmann: yeah but Core also loads by name sometimes: https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs#L208

PR #61, opened 2020-06-19
Configure JSON as default serializer

timbussmann:
Configures JSON as the default serializer, using the NServiceBus.Newtonsoft.Json package as a new dependency. We concluded that the vast majority of users will be using JSON and we can eliminate noisy boiler-plate code for the mandatory serializer configuration by using JSON as a useful default.
The serializer can still be changed, this only introduces an additional package dependency to NServiceBus.Newtonsoft.Json, Newtonsoft.JSON is already a transitive dependency due to the Microsoft.Azure.WebJobs packages.

For the future, it would be more appropriate to use the built-in JSON serializer from .NET Core which should probably not be a major problem as community packages are already making use of this. However, there is no clear understanding of potential compatibility issues so this is scoped out for now. A dedicated issue will be raised to handle the adoption of the native JSON serializer.

### DIFF
--- a/src/NServiceBus.AzureFunctions.ServiceBus/NServiceBus.AzureFunctions.ServiceBus.csproj
+++ b/src/NServiceBus.AzureFunctions.ServiceBus/NServiceBus.AzureFunctions.ServiceBus.csproj
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


timbussmann:
added the dependency range to the NServiceBus.Newtonsoft.Json package given it's the standard on how we treat dependencies. The package has a dependency range on Newtonsoft.JSON anyway.

PR #62, opened 2020-06-19
Saga acceptance test

timbussmann:
Adds an acceptance tests using a saga in a function.
I had to adjust the acceptance testing a little bit:
* similar to core's endpoint tests, only include types in the test class for scanning to avoid issues with missing persistence configuration in other tests
* changed the API a bit to allow sending multiple messages to a function. Because we can completely await the processing of the message, this is fairly easy.

I've removed inmemory persistence as the default configured persistence as this would cause problems with the DI container when the user decides to configure a different saga persistence (which is what we would recommend). There doesn't seem to be a need to configure a default persisted with ASB transport, as ASB brings in many native capabilities that would otherwise require a persisted. Only when sagas are used, a persister configuration becomes necessary.

additional thoughts:
* configuring a saga persister requires the use of the `AdvancedConfiguration` API. Given Sagas might be a bit of an experiment edge case, that might be sufficient. Otherwise, we should consider pulling the persistence config API up to the top-level API too (similar to the serializer)

### DIFF
--- a/src/ServiceBus.Tests/When_using_sagas.cs
+++ b/src/ServiceBus.Tests/When_using_sagas.cs
@@ -71,6 +71,7 @@
                 public Task Handle(ReadSagaDataValueMessage message, IMessageHandlerContext context)
                 {
                     testContext.CounterValue = Data.SomeCounter;
+                    MarkAsComplete();
                     return Task.CompletedTask;
                 }
             }
--- a/src/ServiceBus.Tests/When_using_sagas.cs
+++ b/src/ServiceBus.Tests/When_using_sagas.cs
@@ -28,7 +28,6 @@
         {
             public SendingFunction()
             {
-                //TODO use learning persistence?
                 CustomizeConfiguration = configuration => 
                     configuration.AdvancedConfiguration.UsePersistence<LearningPersistence>();
 
--- a/src/ServiceBus.Tests/When_using_sagas.cs
+++ b/src/ServiceBus.Tests/When_using_sagas.cs
@@ -1,0 +1,102 @@
+﻿namespace ServiceBus.Tests
+{
+    using System;
+    using System.Threading.Tasks;
+    using NServiceBus;
+    using NServiceBus.AcceptanceTesting;
+    using NUnit.Framework;
+
+    public class When_using_sagas
+    {
+        [Test]
+        public async Task Should_invoke_saga_message_handlers()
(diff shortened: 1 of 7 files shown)


src/ServiceBus.Tests/When_using_sagas.cs
```diff
@@ -0,0 +1,102 @@
...
+        class Context : ScenarioContext
+        {
+            public int CounterValue { get; set; }
+        }
+
+        class SendingFunction : FunctionEndpointComponent
+        {
+            public SendingFunction()
+            {
+                //TODO use learning persistence?
```
SeanFeldman, line 31: Remove?
```suggestion
```

src/ServiceBus.Tests/When_using_sagas.cs
```diff
@@ -0,0 +1,102 @@
...
+                CustomizeConfiguration = configuration => 
+                    configuration.AdvancedConfiguration.UsePersistence<LearningPersistence>();
+
+                var correlationProperty = Guid.NewGuid().ToString("N");
+                Messages.Add(new StartSagaMessage {CorrelationProperty = correlationProperty});
+                Messages.Add(new UpdateSagaMessage {CorrelationProperty = correlationProperty, UpdateValue = 42});
```
SeanFeldman, line 37: 🖖

src/ServiceBus.Tests/When_using_sagas.cs
```diff
@@ -0,0 +1,102 @@
...
+
+                public Task Handle(UpdateSagaMessage message, IMessageHandlerContext context)
+                {
+                    Data.SomeCounter = message.UpdateValue;
+                    return Task.CompletedTask;
+                }
+
+                public Task Handle(ReadSagaDataValueMessage message, IMessageHandlerContext context)
+                {
+                    testContext.CounterValue = Data.SomeCounter;
```
SeanFeldman, line 74: For completeness, should we include `MarkAsComplete()`?
```suggestion
                    testContext.CounterValue = Data.SomeCounter;
                    MarkAsComplete();
```

timbussmann: should make a big difference, the saga is stored as a file and afaik we cleanup the whole folders anyway after the test runs. But it doesn't hurt to use the APIs 👍

PR #65, opened 2020-06-25
Change webjobs ASB dependency version from 4.1.2 to 4.1.0

timbussmann:
I noticed that creating a new function project in VS used version 4.1.0 of the `Microsoft.Azure.WebJobs.Extensions.ServiceBus` package. This causes nugget errors because our package requires version 4.1.2+. But since our dependency is transitive and further down in the dependency chain, nuget will use 4.1.0 and warn about a package downgrade. To resolve this, users would have to explicitly update the package to 4.1.2 to make the nuget error go away.

Should be ported to ASQ as well if merged.


timbussmann:
manually triggered the CLA assistant check as it seemed to be stuck

PR #68, opened 2020-06-30
SendsAtomicWithReceive support

SeanFeldman:
Fixes https://github.com/Particular/NServiceBus.AzureFunctions.ServiceBus/issues/43

**WARNING**: experimental and needs to be validated


src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
```diff
@@ -27,49 +30,142 @@ public FunctionEndpoint(Func<FunctionExecutionContext, ServiceBusTriggeredEndpoi
         /// <summary>
         /// Processes a message received from an AzureServiceBus trigger using the NServiceBus message pipeline.
         /// </summary>
-        public async Task Process(Message message, ExecutionContext executionContext, ILogger functionsLogger = null)
+        public async Task Process(Message message, ExecutionContext executionContext, ILogger functionsLogger = null, IMessageReceiver messageReceiver = null)
```
timbussmann, line 33: I think we should consider creating a dedicated overload where the message receiver is not optional (and maybe even name it differently)?

src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
```diff
@@ -27,49 +30,142 @@ public FunctionEndpoint(Func<FunctionExecutionContext, ServiceBusTriggeredEndpoi
...
         {
             FunctionsLoggerFactory.Instance.SetCurrentLogger(functionsLogger);
 
-            var messageContext = CreateMessageContext(message);
+            // TODO: get the transaction mode the endpoint is configured with
+            var transportTransactionMode = TransportTransactionMode.SendsAtomicWithReceive;
+            var useTransaction = messageReceiver != null && transportTransactionMode == TransportTransactionMode.SendsAtomicWithReceive;
```
timbussmann, line 39: why do we need this at all? This is a setting for the message pump which we don't use. If you provide a message receiver, wouldn't that mean you `useTransaction` to be true?

src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
```diff
@@ -27,49 +30,142 @@ public FunctionEndpoint(Func<FunctionExecutionContext, ServiceBusTriggeredEndpoi
...
-                    .ConfigureAwait(false);
-
-                if (errorHandleResult == ErrorHandleResult.Handled)
+                try
+                {
+                    await messageReceiver.SafeDeadLetterAsync(transportTransactionMode, lockToken, deadLetterReason: "Poisoned message", deadLetterErrorDescription: exception.Message).ConfigureAwait(false);
+                }
+                catch (Exception)
+                {
+                    // nothing we can do about it, message will be retried
```
timbussmann, line 62: we could log though?

src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
```diff
@@ -27,49 +30,142 @@ public FunctionEndpoint(Func<FunctionExecutionContext, ServiceBusTriggeredEndpoi
...
+
+                return;
+            }
+
+            try
+            {
+                var transportTransaction = CreateTransportTransaction(useTransaction, messageReceiver, message.PartitionKey);
+                var messageContext = CreateMessageContext(message, messageId, headers, body, transportTransaction);
+
+                using (var scope = useTransaction ? new TransactionScope(TransactionScopeOption.RequiresNew, TransactionScopeAsyncFlowOption.Enabled) : null)
```
timbussmann, line 73: what is the TransactionScope required for? Does ASB use that somehow?

src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
```diff
@@ -27,49 +30,142 @@ public FunctionEndpoint(Func<FunctionExecutionContext, ServiceBusTriggeredEndpoi
...
+                {
+                    await Process(messageContext, functionExecutionContext).ConfigureAwait(false);
+
+                    // Azure Functions auto-completion would be disabled if we try to run in SendsAtomicWithReceive, need to complete message manually
+                    if (useTransaction)
+                    {
+                        await messageReceiver.CompleteAsync(message.SystemProperties.LockToken).ConfigureAwait(false);
+                    }
+
+                    scope?.Complete();
```
timbussmann, line 83: is the message receiver enlisted with the scope?

src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
```diff
@@ -27,49 +30,142 @@ public FunctionEndpoint(Func<FunctionExecutionContext, ServiceBusTriggeredEndpoi
...
-                    new TransportTransaction(),
-                    message.SystemProperties.DeliveryCount);
-
-                var errorHandleResult = await ProcessFailedMessage(errorContext, functionExecutionContext)
-                    .ConfigureAwait(false);
-
-                if (errorHandleResult == ErrorHandleResult.Handled)
+                try
+                {
+                    await messageReceiver.SafeDeadLetterAsync(transportTransactionMode, lockToken, deadLetterReason: "Poisoned message", deadLetterErrorDescription: exception.Message).ConfigureAwait(false);
```
timbussmann, line 58: why wouldn't we just put this in the regular error queue?

src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
SeanFeldman: Good point. I'll change it to be based on the existence of the message receiver.

src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
SeanFeldman: What benefits will it provide?

src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
SeanFeldman: Correct. Needed for ASB's transactional messaging.

src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
SeanFeldman: yes

src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
SeanFeldman: Explained offline. Dead-letter messages cannot be actioned as the messages ending up in the error queue for any other reason. Also, ASB has native support for DLQ which we're leveraging.

src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
timbussmann: better indication to the user what it does to provide this parameter

src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
```diff
@@ -27,49 +29,143 @@ public FunctionEndpoint(Func<FunctionExecutionContext, ServiceBusTriggeredEndpoi
...
+                }
+                catch (Exception ex)
+                {
+                    // nothing we can do about it, message will be retried
+                    functionExecutionContext.Logger.LogDebug($"Failed to move a poisonous message with native ID: `{message.MessageId}` to the dead-letter queue. Message will be retried.", ex);
+                }
+
+                return;
```
timbussmann, line 65: if we would rethrow here instead, what would happen given that we moved the message to the DLQ already?

src/NServiceBus.AzureFunctions.ServiceBus/MessageReceiverExtensions.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace NServiceBus.AzureFunctions.ServiceBus
+{
+    using System.Threading.Tasks;
+    using Microsoft.Azure.ServiceBus.Core;
+
+    static class MessageReceiverExtensions
+    {
+        public static Task SafeCompleteAsync(this IMessageReceiver messageReceiver, TransportTransactionMode transportTransactionMode, string lockToken)
+        {
+            if (transportTransactionMode != TransportTransactionMode.None)
```
timbussmann, line 10: that will never be the case because we only switch between SendOnly and SendsAtomicWithReceive?

src/NServiceBus.AzureFunctions.ServiceBus/MessageReceiverExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+            {
+                return messageReceiver.CompleteAsync(lockToken);
```
timbussmann, line 12: wouldn't that currently throw NREs when not providing a message receiver?

src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
```diff
@@ -27,49 +29,143 @@ public FunctionEndpoint(Func<FunctionExecutionContext, ServiceBusTriggeredEndpoi
...
+
+                        scope?.Complete();
+                    }
+
+                    if (result == ErrorHandleResult.RetryRequired)
+                    {
+                        await messageReceiver.SafeAbandonAsync(transportTransactionMode, lockToken).ConfigureAwait(false);
+
+                        // Indicate to the Functions runtime not to complete the incoming message
+                        throw;
```
timbussmann, line 113: this part should probably not be included in the try/catch because it would be just rethrown again on line 126 with an incorrect log message?

mikeminutillo:
I had a look through this and it looks like it's on the right track. We might want to consider dropping the transaction scope in favor of [explicitly floating a commitable transaction](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/178).

SeanFeldman:
> We might want to consider dropping the transaction scope in favor of explicitly floating a commitable transaction.

Have to, not might want 🙂
This spike PR was bears on the transport prior to the changes that happened how transient transaction is avoided. Hence, it would be required to update the code to use a commitable transaction.

PR #75, opened 2020-07-09
Invoke transport's infrastructure Start and Stop methods

SeanFeldman:
Invoke ASB transport infrastructure `Start()` and `Stop()` methods.
ASB's `Stop()` method is responsible for [closing message senders pool](https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/ed0cc09e334f18b9412606f6c9c236b6848af299/src/Transport/AzureServiceBusTransportInfrastructure.cs#L181-L187) which is used by ASB Functions for dispatching messages.


timbussmann:
I think we never call stop anyway from an azure function endpoint (because there isn't an explicit stop notification we can use, is there?), so it shouldn't really change a thing, but I guess it makes sense anyway.

PR #103, opened 2020-11-04
Support IServiceCollection provided by Functions

SeanFeldman:


timbussmann:
merging this as this just moves the spike to a branch on our repo where we can further adjust & fix.

PR #104, opened 2020-11-05
Exclude Microsoft.Data.SqlClient from assembly scanning

timbussmann:
Scanning the Microsoft.Data.SqlClient assembly in the bin folder causes a "platform not supported" exception. Since we have no need to scan this assembly, excluding it from the scanning process should solve the issue and e.g. allow to use Entity Framework Core with the NServiceBus Azure Functions package


SeanFeldman (approved):
We should pr this against core as well.

SeanFeldman:
[PR against Core](https://github.com/Particular/NServiceBus/pull/5832) was raised.

PR #105, opened 2020-11-05
Support FunctionsHostBuilder configuration

timbussmann:
Add support to register/configure an NServiceBus functions endpoint via the `IFunctionsHostBuilder` API provided by `FunctionsStartup`. This approach gives access to a functions host managed DI container which means we need to support externally managed container mode to integrate properly with the provided ServiceCollection and ServiceProvider.


src/NServiceBus.AzureFunctions.ServiceBus/FunctionsHostBuilderExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Provides extension methods to configure a <see cref="FunctionEndpoint"/> using <see cref="IFunctionsHostBuilder"/>.
+    /// </summary>
+    public static class FunctionsHostBuilderExtensions
+    {
+        /// <summary>
+        /// Configures an NServiceBus endpoint that can be injected into a function trigger as a <see cref="FunctionEndpoint"/> via dependency injection.
+        /// </summary>
+        public static void UseNServiceBus(
+            this IFunctionsHostBuilder functionsHostBuilder,
+            Func<ServiceBusTriggeredEndpointConfiguration> configurationFactory)
```
timbussmann, line 18: should we change this to just take a `ServiceBusTriggeredEndpointConfiguration` instance instead?

e.g. instead of `builder.UseNServiceBus(() => new ServiceBusTriggeredEndpointConfiguration(...));` it would be just
`builder.UseNServiceBus(new ServiceBusTriggeredEndpointConfiguration());`

However, if you need to configure it some more, it might be easier with the function as it feels more like an "inlined" configuration:

```
builder.UseNServiceBus(() => {
    var cfg = new ServiceBusTriggeredEndpointConfiguration(...);
    cfg.SomethingSomething()
    return cfg;
});
```

vs.

```
var cfg = new ServiceBusTriggeredEndpointConfiguration(...);
cfg.SomethingSomething();

builder.UseNServiceBus(cfg);
```

SeanFeldman: Lambda approach (inline) is what customers are used to with other libraries. We should prefer it over `builder.UseNServiceBus(cfg)`.

SeanFeldman (approved):

Issue #113, opened 2020-12-12
No option to mock FunctionEndpoint 


Claker:
Hi, I am trying to unit test my azure function which uses a FunctionEndpoint.
I did not find any option to mock it so I can pass it to the ctor of my new function instance.

This github repo does not have the option to request a feature so I created it as an issue.
Can you please advice?

        private readonly FunctionEndpoint _endpoint;

        public Function(FunctionEndpoint endpoint)
        {
            _endpoint = endpoint;
        }

        [FunctionName(NServiceBusConstants.CreateUserEndpoint)]
        public async Task Run([ServiceBusTrigger(NServiceBusConstants.CreateUserEndpoint)]
            Message message, ILogger logger, ExecutionContext executionContext)
        {
            await _endpoint.Process(message, executionContext, logger);
        }

danielmarbach:
@Claker I have transferred the issue into the corresponding repository. I guess a change like

https://github.com/Particular/NServiceBus.AzureFunctions.ServiceBus/pull/114

would enable the scenarios you described right?

Claker:
Morning @danielmarbach,

Yes! Awesome! 😄 
Thanks for taking the feedback on board and adding a fix !

Claker:
Hi @danielmarbach, do you think the PR will be merged and released early next year?
BTW, have a great Holiday!

SeanFeldman:
@Claker, yes. The plan is to address it after the holidays, early 2021.

danielmarbach:
@Claker I have released it as part of 0.4.0 and updated the documentation. The official release announcement will soon be posted to the discuss group. Thanks for your patience while I was eating way too many Christmas cookies.

danielmarbach:
Here is the announcement link

https://discuss.particular.net/t/nservicebus-azurefunctions-azureservicebus-0-4-0-minor-release-available/2235

Claker:
Thank you very much!!
The company I work at (Serko Ltd.) are currently preparing to buy the licenses we need to use NServiceBus in Prod and we might use it a lot more in the future. 
We are still experimenting with it in Dev and this is the first time we try it.

If we have more questions, what would be a good channel to ask? Would github be ok or your website or is there another recommended way?

danielmarbach:
@Claker All our available support channels can be found on https://particular.net/support

[...]

PR #114, opened 2020-12-15
Introduce interface for the function endpoint to simplify testing

danielmarbach:
Solves #113 

Given this is a preview it would be fine to release this as 0.3 right? If you agree with the changes I can also make the [same change for ASQ](https://github.com/Particular/NServiceBus.AzureFunctions.StorageQueues/pull/50)

Alternatively, we could make the Process method virtual but I think it would be more invasive because we would need to make sure production use cases off overriding those methods would be supported.

### DIFF
--- a/src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
+++ b/src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
@@ -18,7 +18,7 @@
     /// An NServiceBus endpoint hosted in Azure Function which does not receive messages automatically but only handles
     /// messages explicitly passed to it by the caller.
     /// </summary>
-    public class FunctionEndpoint
+    public class FunctionEndpoint : IFunctionEndpoint
     {
         /// <summary>
         /// Creates a new instance of <see cref="FunctionEndpoint" /> that can handle messages using the provided configuration.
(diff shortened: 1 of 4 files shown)


timbussmann:
is there some kind of test that we can add to ensure that `IFunctionEndpoint` and `FunctionEndpoint` is correctly resolvable via ctor injection?

danielmarbach:
@timbussmann at this stage I did not change the DI mechanism. So it would not be possible to inject `IFunctionEndpoint` just yet but I guess you are right it would make sense to align that as well. I guess if I add support for that as well it might increase the scope because it would require an upgrade guide in case only DI via `IFunctionEndpoint` would be supported unless we do as you said and support both.

danielmarbach:
@timbussmann unfortunately the host builder extensions are not tested. So I added code that enables to inject both

timbussmann:
> So it would not be possible to inject IFunctionEndpoint just yet but I guess you are right it would make sense to align that as well.

I think I'd kinda assume that this is supported. Given the test scenario mentioned in the linked issue, it would even be necessary wouldn't it?

supporting ctor injection of the interface or the specific type seems to be the best approach at this point?

danielmarbach:
@timbussmann see https://github.com/Particular/NServiceBus.AzureFunctions.ServiceBus/pull/114/commits/5cb492ad92ff1dad580a39743f2ac98fc591c3b7#diff-788e3caf43645f82e01b7e5619c0010c715d4fbb1413ddf89873de64e65ecde9R25-R27

SeanFeldman:
> Given this is a preview it would be fine to release this as 0.3 right?

0.4, 0.3 is the [currently released](https://www.nuget.org/packages/NServiceBus.AzureFunctions.ServiceBus/) preview.

danielmarbach:
Sure. Thanks @SeanFeldman I missed that.

timbussmann (approved):

PR #115, opened 2021-01-07
Introduce support for operations outside of message handler

### LINKED ISSUE #82: Send message outside of handler
We have a few cases where we send a message at endpoint startup to do things like self register or self initialize a saga. I would like to be able to send a message when the FunctionEndpoint is created, but not in the context of the message being handled. Would it be possible to have FunctionEndpoint expose a getter for iMessage session?

SeanFeldman:
Closes #82 

With this change, messages can be sent/published outside of the message handler, similar to what can be done with `IEndpointInstance` or `IMessageSession`. This enables scenarios such as dispatching messages from Timer and HTTP triggers, using both the static and `FunctionsHostBuilder` approach.

Using `FunctionsHostBuilder` approach with `HttpTrigger`:
image

Dispatching a message on `TimerTrigger`-ed Function:
image


boblangley:
Should `IFunctionEndpoint` now also extend `IMessageSession` ?

SeanFeldman:
> Should `IFunctionEndpoint` now also extend `IMessageSession` ?

That's how I initially started, with extending `IFunctionEndpoint` with `IMessageSession`.
The downside of that approach was that `IMessageSession` is not designed for Functions runtime and `ExecutionContext` and `ILogger` are not part of the signature. Therefore `Send` and `Publish` w/o those two would be possible but not desirable. I'm also not convinced that having `IMessageSession` would add much value as DI would provide `IFunctionsEndpoint`. And it's consistent across the Functions implementation with a smaller public API surface.

For these reasons, I think trying to introduce `IMessageSession` is not needed.

boblangley (approved):

PR #128, opened 2021-02-05
V8 transport seam

timbussmann:
Update to the new seam. Since the subscription manager is tied to a receiver, we need to initialize the whole receiver (but not start it).


src/NServiceBus.AzureFunctions.ServiceBus/NServiceBus.AzureFunctions.ServiceBus.csproj
```diff
@@ -9,8 +9,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.Functions.Extensions" Version="1.1" />
     <PackageReference Include="NServiceBus.Newtonsoft.Json" Version="2.2.0" />
-    <PackageReference Include="NServiceBus.Transport.AzureServiceBus" Version="1.8.0" />
-    <PackageReference Include="NServiceBus.Extensions.DependencyInjection" Version="1.0.1" />
```
SeanFeldman, line 13: This is now baked into Core, right? 😍

src/NServiceBus.AzureFunctions.ServiceBus/Serverless/TransportWrapper/ServerlessTransport.cs
```diff
@@ -1,26 +1,43 @@
...
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
     using Transport;
 
     class ServerlessTransport<TBaseTransport> : TransportDefinition
-        where TBaseTransport : TransportDefinition, new()
+        where TBaseTransport : TransportDefinition
     {
-        public ServerlessTransport()
+        const string MainReceiverId = "Main"; // the fixed ID of the main receiver, assigned by core.
```
SeanFeldman, line 10: > // the fixed ID of the main receiver, assigned by core.

Core doesn't provide the constant? Like Headers, I'd expect to see that value provided by Core.

SeanFeldman (commented):
A few minor comments.
There's one test that is consistently failing (Windows and Linux)

src/NServiceBus.AzureFunctions.ServiceBus/NServiceBus.AzureFunctions.ServiceBus.csproj
timbussmann: yes 🎉

src/NServiceBus.AzureFunctions.ServiceBus/Serverless/TransportWrapper/ServerlessTransport.cs
timbussmann: no, that constant is currently not exposed publicly. Might be something to consider.

src/ServiceBus.Tests/FunctionEndpointComponent.cs
```diff
@@ -92,7 +94,11 @@ public override Task Start(CancellationToken token)
                                 return Task.CompletedTask;
                             }));
 
-                    endpointConfiguration.RegisterComponents(c => c.RegisterSingleton(scenarioContext.GetType(), scenarioContext));
+                    endpointConfiguration.RegisterComponents(c => c.AddSingleton(scenarioContext.GetType(), scenarioContext));
+
+                    // enable installers to auto-create the input queue for tests
+                    // in real Azure Functions, the input queue is assumed to exist
+                    endpointConfiguration.EnableInstallers();
```
timbussmann, line 101: this is now necessary because the serverless endpoints will always try to create the subscription for the endpoint, which will fail the tests because we didn't actually create the input queue for the function component. The assumption is, that the queue must be there anyway for the function to startup, so we only need to ensure the queue will be created for tests. Do you see any issues with that @SeanFeldman ?

src/NServiceBus.AzureFunctions.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
```diff
@@ -45,26 +44,30 @@ public ServiceBusTriggeredEndpointConfiguration(string endpointName, string conn
...
+            serverlessTransport = new ServerlessTransport<AzureServiceBusTransport>(serviceBusTransport);
+            Routing = EndpointConfiguration.UseTransport(serverlessTransport);
 
             var recoverability = AdvancedConfiguration.Recoverability();
             recoverability.Immediate(settings => settings.NumberOfRetries(5));
             recoverability.Delayed(settings => settings.NumberOfRetries(3));
 
             EndpointConfiguration.UseSerialization<NewtonsoftSerializer>();
+
+            //TODO should we disable Autosubscribe?
```
tmasternak, line 61: Why should we disable it?

tmasternak (approved):

src/NServiceBus.AzureFunctions.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
timbussmann: technically it doesn't make much sense to autosubscribe as the function endpoint only runes once a message has been received. I'm not sure whether it is a valid strategy to have a specific message arrive at the function and only then let the function endpoint subscribe to events, rather than setting up all subscriptions upfront manually (see https://docs.particular.net/previews/azure-functions-service-bus#preparing-the-azure-service-bus-namespace-subscribing-to-events)

tmasternak: Ahh. That makes sense. 👍 for manual subsription.

timbussmann: thoughts @SeanFeldman ☝️ ?

SeanFeldman: I'm on a fence. When using OOTB Functions, the infrastructure has to be provisioned up-front. But many prefer to handle the infrastructure along with Function creation. There's undeniable [evidence ](https://github.com/Particular/NServiceBus.AzureFunctions.ServiceBus/issues/94)that some of the existing customers are already asking for it. I even took the idea a step further and built a quick [spike](https://weblogs.asp.net/sfeldman/automatic-nservicebus-topology-creation-for-function) to help those that wanted the feature already.

Therefore I would not recommend disabling auto-subscription.

src/ServiceBus.Tests/FunctionEndpointComponent.cs
SeanFeldman: As long as the tests are creating the queue, don't have any issues with the approach.

src/NServiceBus.AzureFunctions.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
SeanFeldman, line 61:
```suggestion
```

SeanFeldman (approved):
Approved with a minor suggestion

PR #141, opened 2021-02-22
Remove the assembly scanning workaround

timbussmann:
Removes the workaround as core allows specifying additional scanning directories.

~~Note, this doesn't work currently due to Core also scanning the functions tooling folder, causing conflicts with already loaded assemblies.~~ The following [commit](https://github.com/Particular/NServiceBus.AzureFunctions.ServiceBus/pull/141/commits/f77d99b1511d09c70569ae8e532b17397eb3d7e8) fixes it.

### DIFF
--- a/src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
+++ b/src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
@@ -2,8 +2,6 @@
 {
     using System;
     using System.IO;
-    using System.Reflection;
-    using System.Runtime.Loader;
     using System.Threading;
     using System.Threading.Tasks;
     using AzureFunctions.ServiceBus;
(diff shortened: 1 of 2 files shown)


src/NServiceBus.AzureFunctions.ServiceBus/FunctionEndpoint.cs
```diff
@@ -27,9 +25,9 @@ public FunctionEndpoint(Func<FunctionExecutionContext, ServiceBusTriggeredEndpoi
         {
             endpointFactory = executionContext =>
             {
-                LoadAssemblies(AssemblyDirectoryResolver(executionContext));
-
                 configuration = configurationFactory(executionContext);
+                configuration.EndpointConfiguration.AssemblyScanner().AdditionalAssemblyScanningPath =
```
SeanFeldman, line 29: What's the reason to specify an additional assembly scanning path if it's the same as the AsseblyContext path used by Core?

timbussmann: that's not the path used by core. Core only scans the AppDomain base directory which is the azure functions tooling folder

timbussmann:
@SeanFeldman your latest change is exactly the reason I raised this PR in its original way and just excluding conflicting assemblies should not be the solution here. The workaround should not be removed until the core issue is fixed.

SeanFeldman:
This PR would need to target this repo when the `master` is on Core v8.
Closing for now.

PR #174, opened 2021-04-09
Rename to NServiceBus.AzureFunctions.InProcess.ServiceBus

yvesgoeleven:


SeanFeldman:
Replacing with a newer PR.

PR #176, opened 2021-04-09
Finalize public API

SeanFeldman:


SeanFeldman:
Master was updated to Core v8 while GA requires Core v7. This work will need to be done once `master` is back on Core v7.

PR #235, opened 2021-07-27
Do not require connection string environment variable

### LINKED ISSUE #232: Connection string environment variable is required even if it is not used
This package loads a connection string from an environment variable which by default is named `AzureWebJobsServiceBus` (but this can be overridden in the constructor for `ServiceBusTriggeredEndpointConfiguration`).

This is not the only way to set the transport connection. It can also be set like this:

```csharp
var endpointConfiguration = new ServiceBusTriggeredEndpointConfiguration("MyEndpoint");
endpointConfiguration.Transport.ConnectionString(myConnectionString);
```

Unfortunately this fails as the endpoint still attempts to load the connection string from an environment variable, even if there is one. [This code](https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/blob/master/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs#L50-L52) attempts to set the connection string to `null` which results in an exception:

> NServiceBus.Core: Value cannot be null. (Parameter 'connectionString').

## Workaround

To workaround this problem, a connection string must be available in the Azure Function environment variables and the `ServiceBusTriggeredEndpointConfiguration` should be configured to use this environment variable. Note that the connection string does not need to be for a valid ServiceBus account as it will not be used.

## Expected Behavior

It should be possible to set the connection string on the transport without setting an environment variable.

mikeminutillo:
Resolves #232

If there are functions with a Service Bus trigger then the Azure Functions infrastructure requires that the Service Bus connection string is available in an environment variable, and the NServiceBus endpoint must use this same connection string.

If there are no functions with a Service Bus trigger then the environment variable is not required and the NServiceBus endpoint connection string can be configured using an alternative mechanism. For example:

```csharp
class Startup : FunctionsStartup
{
    public override void Configure(IFunctionsHostBuilder builder)
    {
        var configuration = builder.GetContext().Configuration;
        builder.UseNServiceBus(() =>
        {
            var config = new ServiceBusTriggeredEndpointConfiguration("MyEndpoint");
            config.Transport.ConnectionString(configuration.GetConnectionString("MyConnection"));
            return config;
        });
    }
}
```

### DIFF
--- a/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
+++ b/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
@@ -49,7 +49,15 @@
 
             var connectionString =
                 Environment.GetEnvironmentVariable(connectionStringName ?? DefaultServiceBusConnectionName);
-            Transport.ConnectionString(connectionString);
+            if (!string.IsNullOrWhiteSpace(connectionString))
+            {
+                Transport.ConnectionString(connectionString);
+            }
+            else if (!string.IsNullOrWhiteSpace(connectionStringName))
+            {
+                throw new Exception(
+                    $"Connection string not found. Make sure the Service Bus connection string is stored in an environment variable named {connectionStringName}.");
+            }
 
             var recoverability = AdvancedConfiguration.Recoverability();
             recoverability.Immediate(settings => settings.NumberOfRetries(5));
--- a/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
+++ b/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
@@ -56,7 +56,7 @@
             else if (!string.IsNullOrWhiteSpace(connectionStringName))
             {
                 throw new Exception(
-                    $"Connection string not found. Make sure the Service Bus connection string is stored in an environment variable named {connectionStringName}.");
+                    $"Azure Service Bus connection string named '{connectionStringName}' was provided but wasn't found in the environment variables. Make sure the connection string is stored in the environment variable named '{connectionStringName}'.");
             }
 
             var recoverability = AdvancedConfiguration.Recoverability();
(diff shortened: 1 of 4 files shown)


src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
```diff
@@ -49,7 +49,15 @@ public ServiceBusTriggeredEndpointConfiguration(string endpointName, string conn
...
                 Environment.GetEnvironmentVariable(connectionStringName ?? DefaultServiceBusConnectionName);
-            Transport.ConnectionString(connectionString);
+            if (!string.IsNullOrWhiteSpace(connectionString))
+            {
+                Transport.ConnectionString(connectionString);
+            }
+            else if (!string.IsNullOrWhiteSpace(connectionStringName))
+            {
+                throw new Exception(
+                    $"Connection string not found. Make sure the Service Bus connection string is stored in an environment variable named {connectionStringName}.");
```
SeanFeldman, line 59:
```suggestion
                    $"Azure Service Bus connection string named '{connectionStringName}' was provided but wasn't found in the environment variables. Make sure the connection string is stored in the environment variable named '{connectionStringName}'.");
```

SeanFeldman (approved):
A minor tweak to the exception message.

mikeminutillo:
This is the error message you get if _no_ connection string is provided at all:

> [2021-07-27T05:05:41.959Z] A host error has occurred during startup operation '21fa81c9-fbd2-4357-970d-fb1c8ae38ce5'.
> [2021-07-27T05:05:41.960Z] NServiceBus.Core: Transport connection string has not been explicitly configured via 'ConnectionString' method. Here is an example of what is required: endpointConfig.UseTransport<ServerlessTransport`1>().ConnectionString("");.
> Value cannot be null. (Parameter 'provider')

In a local development environment this error is repeated (the functions startup execution is getting retried)

image

SeanFeldman:
Transport connection string is mandatory with NServiceBus Core v7, right?

mikeminutillo:
> Transport connection string is mandatory with NServiceBus Core v7, right?

Yes. [If the transport definition says it is required then core throws an exception if it's missing](https://github.com/Particular/NServiceBus/blob/7.5.0/src/NServiceBus.Core/Unicast/Transport/TransportConnectionString.cs#L40).

I went looking to see if we can detect a missing connection string before this and throw a better exception but I don't think it's possible without some messy reflection.

HEskandari (approved):

PR #236, opened 2021-07-27
Use IConfiguration to setup endpoint

### LINKED ISSUE #95: Allow endpoint configuration to use IConfiguration
In my Azure Function I am heavily using the `IConfiguration`. More specifically, my connection string to Azure Service Bus is accessed via App Config Service, which in turns gets the value from Key Vault. Ultimately, my actual connection string is available in the `IConfiguration`.  However, based on what I'm seeing it appears that NSB must load the configuration string from an environment variable. See [here](https://github.com/Particular/NServiceBus.AzureFunctions.ServiceBus/blob/b0b734fefa825983bb5d384100bcd71a24a6bb34/src/NServiceBus.AzureFunctions.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs#L32). Ideally, I'd like to pass you a settings 'key' and the value is fetched from an `IConfiguration`.

cc @SeanFeldman

HEskandari:
Closes #95 

More broadly uses IConfiguration to set up the endpoint.

```
class Startup : FunctionsStartup
{
    public override void Configure(IFunctionsHostBuilder builder)
    {
        builder.UseNServiceBus();
    }
}
```

Will need to apply on top of https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/pull/235 when it is merged.

### DIFF
--- a/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionsHostBuilderExtensions.cs
+++ b/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionsHostBuilderExtensions.cs
@@ -3,6 +3,7 @@
     using System;
     using System.IO;
     using Microsoft.Azure.Functions.Extensions.DependencyInjection;
+    using Microsoft.Extensions.Configuration;
     using Microsoft.Extensions.DependencyInjection;
 
     /// <summary>
@@ -10,6 +11,16 @@
     /// </summary>
     public static class FunctionsHostBuilderExtensions
     {
+        /// <summary>
+        /// Use the IConfiguration to configures an NServiceBus endpoint that can be injected into a function trigger as a <see cref="FunctionEndpoint"/> via dependency injection.
+        /// </summary>
+        public static void UseNServiceBus(
+            this IFunctionsHostBuilder functionsHostBuilder)
+        {
+            functionsHostBuilder.UseNServiceBus(() => new ServiceBusTriggeredEndpointConfiguration(
+                functionsHostBuilder.GetContext().Configuration));
+        }
+
         /// <summary>
         /// Configures an NServiceBus endpoint that can be injected into a function trigger as a <see cref="FunctionEndpoint"/> via dependency injection.
         /// </summary>
--- a/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionsHostBuilderExtensions.cs
+++ b/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionsHostBuilderExtensions.cs
@@ -3,7 +3,6 @@
     using System;
     using System.IO;
     using Microsoft.Azure.Functions.Extensions.DependencyInjection;
-    using Microsoft.Extensions.Configuration;
     using Microsoft.Extensions.DependencyInjection;
 
     /// <summary>
--- a/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
+++ b/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
@@ -5,6 +5,7 @@
     using AzureFunctions.InProcess.ServiceBus;
     using Logging;
     using Microsoft.Azure.WebJobs;
+    using Microsoft.Extensions.Configuration;
     using Serialization;
     using Transport;
 
@@ -19,9 +20,33 @@
         }
 
         /// <summary>
+        /// Creates a serverless NServiceBus endpoint running within an Azure Service Bus trigger
+        /// </summary>
+        public ServiceBusTriggeredEndpointConfiguration(string endpointName, IConfiguration configuration = null)
+            : this(endpointName, null, configuration)
+        {
+        }
+
+        /// <summary>
         /// Creates a serverless NServiceBus endpoint running within an Azure Service Bus trigger.
         /// </summary>
         public ServiceBusTriggeredEndpointConfiguration(string endpointName, string connectionStringName = null)
+            : this(endpointName, connectionStringName, null)
+        {
+        }
+
+        /// <summary>
+        /// Creates a serverless NServiceBus endpoint running within an Azure Service Bus trigger.
+        /// </summary>
+        public ServiceBusTriggeredEndpointConfiguration(string endpointName)
+            : this(endpointName, null, null)
+        {
+        }
+
+        /// <summary>
+        /// Creates a serverless NServiceBus endpoint running within an Azure Service Bus trigger.
+        /// </summary>
+        internal ServiceBusTriggeredEndpointConfiguration(string endpointName, string connectionStringName = null, IConfiguration configuration = null)
         {
             EndpointConfiguration = new EndpointConfiguration(endpointName);
 
@@ -33,13 +58,12 @@
             EndpointConfiguration.CustomDiagnosticsWriter(_ => Task.CompletedTask);
 
             // 'WEBSITE_SITE_NAME' represents an Azure Function App and the environment variable is set when hosting the function in Azure.
-            var functionAppName = Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME") ?? Environment.MachineName;
+            var functionAppName = GetConfiguredValueOrFallback(configuration, "WEBSITE_SITE_NAME", true) ?? Environment.MachineName;
             EndpointConfiguration.UniquelyIdentifyRunningInstance()
                 .UsingCustomDisplayName(functionAppName)
                 .UsingCustomIdentifier(DeterministicGuid.Create(functionAppName));
 
-            // Look for license as an environment variable
-            var licenseText = Environment.GetEnvironmentVariable("NSERVICEBUS_LICENSE");
+            var licenseText = GetConfiguredValueOrFallback(configuration, "NSERVICEBUS_LICENSE", optional: true);
             if (!string.IsNullOrWhiteSpace(licenseText))
             {
                 EndpointConfiguration.License(licenseText);
@@ -47,16 +71,17 @@
 
             Transport = UseTransport<AzureServiceBusTransport>();
 
-            var connectionString =
-                Environment.GetEnvironmentVariable(connectionStringName ?? DefaultServiceBusConnectionName);
+            var connectionString = GetConfiguredValueOrFallback(configuration, connectionStringName ?? DefaultServiceBusConnectionName, optional: true);
             if (!string.IsNullOrWhiteSpace(connectionString))
             {
                 Transport.ConnectionString(connectionString);
             }
-            else if (!string.IsNullOrWhiteSpace(connectionStringName))
+            else
             {
-                throw new Exception(
-                    $"Azure Service Bus connection string named '{connectionStringName}' was provided but wasn't found in the environment variables. Make sure the connection string is stored in the environment variable named '{connectionStringName}'.");
+                var message = connectionStringName != null ?
+                    $"Azure Service Bus connection string named '{connectionStringName}' was provided but wasn't found in the environment variables. Make sure the connection string is stored in the environment variable named '{connectionStringName}'." :
+                    "Azure Service Bus connection string was not provided.";
+                throw new Exception(message);
             }
 
             var recoverability = AdvancedConfiguration.Recoverability();
@@ -64,6 +89,25 @@
             recoverability.Delayed(settings => settings.NumberOfRetries(3));
 
             EndpointConfiguration.UseSerialization<NewtonsoftSerializer>();
+        }
+
+        string GetConfiguredValueOrFallback(IConfiguration configuration, string key, bool optional)
+        {
+            if (configuration != null)
+            {
+                var configuredValue = configuration.GetValue<string>(key);
+                if (configuredValue != null)
+                {
+                    return configuredValue;
+                }
+            }
+
+            var environmentVariable = Environment.GetEnvironmentVariable(key);
+            if (string.IsNullOrEmpty(environmentVariable) && !optional)
+            {
+                throw new Exception($"Configuration or environment value for '{key}' was not set or was empty.");
+            }
+            return environmentVariable;
         }
 
         /// <summary>
--- a/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
+++ b/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
@@ -84,12 +84,9 @@
             {
                 Transport.ConnectionString(connectionString);
             }
-            else
+            else if (!string.IsNullOrWhiteSpace(connectionStringName))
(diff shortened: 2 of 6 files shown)


src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
```diff
@@ -18,10 +19,26 @@ static ServiceBusTriggeredEndpointConfiguration()
...
+        public ServiceBusTriggeredEndpointConfiguration(string endpointName, string connectionString = null)
+            : this(endpointName, connectionString, null)
+        {
+        }
+
         /// <summary>
         /// Creates a serverless NServiceBus endpoint running within an Azure Service Bus trigger.
         /// </summary>
-        public ServiceBusTriggeredEndpointConfiguration(string endpointName, string connectionStringName = null)
+        internal ServiceBusTriggeredEndpointConfiguration(string endpointName, string connectionString = null, IConfiguration configuration = null)
```
SeanFeldman, line 41: Note that this will be a breaking change and require a new major since connection string != connection string name.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
```diff
@@ -80,7 +115,7 @@ public static ServiceBusTriggeredEndpointConfiguration FromAttributes()
             var configuration = TriggerDiscoverer.TryGet<ServiceBusTriggerAttribute>();
             if (configuration != null)
             {
-                return new ServiceBusTriggeredEndpointConfiguration(configuration.QueueName, configuration.Connection);
+                return new ServiceBusTriggeredEndpointConfiguration(configuration.QueueName, configuration.Connection); //TODO: How to fix this?
```
SeanFeldman, line 118: You can't. The attribute doesn't provide runtime values, only static, which a connection string **name** is.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
HEskandari: I can undo that to keep the old version.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
HEskandari: Should be fixed in the latest commit. I've reverted back to using it as a connection string name.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
```diff
@@ -33,22 +50,21 @@ public ServiceBusTriggeredEndpointConfiguration(string endpointName, string conn
...
             if (!string.IsNullOrWhiteSpace(licenseText))
             {
                 EndpointConfiguration.License(licenseText);
             }
 
             Transport = UseTransport<AzureServiceBusTransport>();
 
-            var connectionString =
-                Environment.GetEnvironmentVariable(connectionStringName ?? DefaultServiceBusConnectionName);
+            var connectionString = GetConfiguredValueOrFallback(configuration, connectionStringName ?? DefaultServiceBusConnectionName);
```
SeanFeldman, line 75: I think this is the scenario where you want to be very explicit and indicate that this configuration is not optional, leading to `bool optional` not having a default value of `false`.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
```diff
@@ -58,6 +74,25 @@ public ServiceBusTriggeredEndpointConfiguration(string endpointName, string conn
             EndpointConfiguration.UseSerialization<NewtonsoftSerializer>();
         }
 
+        private string GetConfiguredValueOrFallback(IConfiguration configuration, string key, bool optional = false)
+        {
+            if (configuration != null)
+            {
+                var configuredValue = configuration.GetValue<string>(key);
+                if (configuredValue != null)
```
SeanFeldman, line 90: null or empty stirng?

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
```diff
@@ -58,6 +74,25 @@ public ServiceBusTriggeredEndpointConfiguration(string endpointName, string conn
...
+                {
+                    return configuredValue;
+                }
+            }
+
+            var environmentVariable = Environment.GetEnvironmentVariable(key);
+            if (environmentVariable == null && !optional)
```
SeanFeldman, line 97: null or empty stirng?

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
```diff
@@ -58,6 +74,25 @@ public ServiceBusTriggeredEndpointConfiguration(string endpointName, string conn
...
+            {
+                throw new InvalidOperationException($"Configuration or environment value for '{key}' was not set.");
```
SeanFeldman, line 99: was not set or was empty

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
HEskandari: I thought about that. I think it will work for now as for all the specific purposes we have non-null or empty string is expected.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
HEskandari: @SeanFeldman not sure I follow, do you mean:

- To change the optional default to false?
- Or pass optional = true for the connection string and throw a more meaningful / explicit exception?

SeanFeldman: I mean not to make `optional` an optional parameter. Rather, be explicit and always pass true or false. It doesn't save much but adds clarity to see what's optional and what's not. Further, I'd also add the parameter name to the call for readability.
E.g.:
```c#
var connectionString = GetConfiguredValueOrFallback(configuration, 
  connectionStringName ?? DefaultServiceBusConnectionName, optional:true);
```

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
HEskandari: done

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
HEskandari: done

mikeminutillo:
Is it worth creating an overload of `UseNserviceBus()` that passes `IConfiguration` in to the ServiceBusTriggeredEndpointConfiguration factory method to prevent the closure on IConfiguration and to make it simpler for customers to access?

Additionally, if the configuration can include the endpoint name, would it be worth creating an overload that accepts no parameters and simply loads everything from configuration?

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
```diff
@@ -18,10 +19,42 @@ static ServiceBusTriggeredEndpointConfiguration()
             LogManager.UseFactory(FunctionsLoggerFactory.Instance);
         }
 
+        /// <summary>
+        /// Creates a serverless NServiceBus endpoint running within an Azure Service Bus trigger
```
mikeminutillo, line 23: These don't _have_ to exist within an Azure Service Bus trigger. Our own documentation includes [an example of running inside an HTTP trigger](https://docs.particular.net/nservicebus/hosting/azure-functions/service-bus#basic-usage-dispatching-outside-a-message-handler).

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
```diff
@@ -33,30 +66,30 @@ public ServiceBusTriggeredEndpointConfiguration(string endpointName, string conn
...
             Transport = UseTransport<AzureServiceBusTransport>();
 
-            var connectionString =
-                Environment.GetEnvironmentVariable(connectionStringName ?? DefaultServiceBusConnectionName);
+            var connectionString = GetConfiguredValueOrFallback(configuration, connectionStringName ?? DefaultServiceBusConnectionName, optional: true);
             if (!string.IsNullOrWhiteSpace(connectionString))
             {
                 Transport.ConnectionString(connectionString);
             }
-            else if (!string.IsNullOrWhiteSpace(connectionStringName))
```
mikeminutillo, line 56: I think this `if` is needed. If we fall back to the `DefaultServiceBusConnectionName` but it's not present, then that's OK. That's exactly what #235 is meant to address. We only throw an exception if you told us to use a specific environment variable but then didn't put anything in it.

src/ServiceBus.Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
```diff
@@ -53,6 +54,9 @@ namespace NServiceBus
     }
     public class ServiceBusTriggeredEndpointConfiguration
     {
+        public ServiceBusTriggeredEndpointConfiguration(Microsoft.Extensions.Configuration.IConfiguration configuration) { }
+        public ServiceBusTriggeredEndpointConfiguration(string endpointName) { }
+        public ServiceBusTriggeredEndpointConfiguration(string endpointName, Microsoft.Extensions.Configuration.IConfiguration configuration = null) { }
```
mikeminutillo, line 59: Can the compiler tell the difference between this constructor and the one above? I'm surprised we don't get an ambiguous method name compiler error.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionsHostBuilderExtensions.cs
```diff
@@ -3,13 +3,24 @@
...
     /// Provides extension methods to configure a <see cref="FunctionEndpoint"/> using <see cref="IFunctionsHostBuilder"/>.
     /// </summary>
     public static class FunctionsHostBuilderExtensions
     {
+        /// <summary>
+        /// Use the IConfiguration to configures an NServiceBus endpoint that can be injected into a function trigger as a <see cref="FunctionEndpoint"/> via dependency injection.
+        /// </summary>
+        public static void UseNServiceBus(
+            this IFunctionsHostBuilder functionsHostBuilder,
+            IConfiguration configuration)
```
timbussmann, line 19: why do we have to pass the IConfiguration here if we can just grab it from the `IFunctionsHostBuilder` directly?

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
HEskandari: Updated

src/ServiceBus.Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
HEskandari: It depends. When calling the constructor, if you do `new ServiceBusTriggeredEndpointConfiguration("test", null)` you do get ambiguous constructor reference, since it can resolve null for both IConfiguration and connectionStringName.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionsHostBuilderExtensions.cs
HEskandari: This is fixed in the latest commit.

src/ServiceBus.Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
mikeminutillo: But if I do `new ServiceBusTriggeredEndpointConfiguration("test")` which constructor gets called? Presumably the one wiithout the nullable parameter. If that's the case, why make this parameter nullable?

mikeminutillo (approved):
Looks good. I proposed an additional overload here #246

src/ServiceBus.Tests/When_no_connection_string_is_provided.cs
```diff
@@ -12,6 +12,9 @@ public class When_no_connection_string_is_provided
         public void Should_guide_user_towards_success()
         {
             var endpointConfiguration = new EndpointConfiguration("SampleEndpoint");
+            var scanner = endpointConfiguration.AssemblyScanner();
+            scanner.ThrowExceptions = false;
```
timbussmann, line 16: why is this necessary here?

HEskandari: For some reason, it is necessary everywhere when using Rider. Basically, nunit.engine.dll is picked up by both test runner and assembly scanner and that causes exceptions.

timbussmann: but it's only configured on this test?
Looking at core, the next major version seems to have the unit.engine dll excluded by default. might be something that we have to configure for the test project if an update of rider doesn't help, but changing this only on a single test sounds off.

Also sounds like something that should be addressed as a separate PR if necessary?

HEskandari: @timbussmann it was done in a few other places that was broken, via a [different PR](https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/pull/234).

timbussmann: we definitely need to make sure we remove this again in the v8 upgrade because core should exclude those by default. Otherwise we're running into risks of not detecting actual issues with the assembly scanning.

timbussmann (approved):

PR #237, opened 2021-07-27
ASB transaction support

### LINKED ISSUE #43: Add support for SendsAtomicWithReceive

timbussmann:
Resolves #43
Based on https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/pull/68.

Adds support for transactional message processing


src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
```diff
@@ -28,14 +30,97 @@ public class FunctionEndpoint : IFunctionEndpoint
...
+
+                        MessageContext messageContext = CreateMessageContext(message, transportTransaction);
+
+                        //TODO: Could also be done earlier since the token here doesn't need to come from the context (it's a new one anyway)
+                        await InitializeEndpointIfNecessary(functionExecutionContext,
+                            messageContext.ReceiveCancellationTokenSource.Token).ConfigureAwait(false);
+
+                        await pipeline.PushMessage(messageContext).ConfigureAwait(false);
+
+                        using (var scope = new TransactionScope(transaction, TransactionScopeAsyncFlowOption.Enabled))
```
SeanFeldman, line 56: Should we bring in the extension methods from the ASB transport to A) make this less overwhelming and B) read more like the transport? I find `SafeComplete` and `SafeAbandon` with the scope suppression logic encapsulated there a bit more readable. It also plays nicely with a single method for both transactional modes, on which I'll expand in a comment below.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/IFunctionEndpoint.cs
```diff
@@ -12,6 +13,11 @@
     /// </summary>
     public interface IFunctionEndpoint
     {
+        /// <summary>
+        /// Processes a message received from an AzureServiceBus trigger using the NServiceBus message pipeline. All messages are committed transactionally with the successful processing of the incoming message.
+        /// </summary>
+        Task ProcessTransactional(Message message, ExecutionContext executionContext, IMessageReceiver messageReceiver, ILogger functionsLogger = null);
```
SeanFeldman, line 19: I would prefer to see a single public `Process` method over two variants.

Wherever the trigger function code is auto-generated or not, NServiceBus will need to know wherever to enable cross-entity transactions or not. The key differentiator between a cross-entity transaction mode (`SendsAtomicWithReceive`) vs a non-transactional mode (`ReceiveOnly`) is the additional argument on the `ServiceBusTrigger` to indicate the incoming message should not be auto-completed by the Functions SDK.

In most scenarios, the method is in the boilerplate trigger function code that is autogenerated. That code could always require `MessageReceiver` to be passed into the `Process` method.  And users opting out of trigger function auto-generation needing to provide that argument. The concern is that there's a chance that the `AutoComplete` argument to the attribute might be forgotten and with a single `Process` method it won't be obvious could be addressed with a code inspection and a mandatory additional argument on the `Process` method, that would be aligned with the attribute's boolean flag for `AutoComplete`. 

Attributes:

```c#
[assembly: NServiceBusEndpointName(name: "MyEndpoint", triggerFunctionName: "MyEndpointTrigger")]
[assembly: NServiceBusEndpointEnableCrossEntityTransactions(bool)]
```

Trigger function:

```c#
class FunctionEndpointTrigger
{
  readonly IFunctionEndpoint endpoint;

  public FunctionEndpointTrigger(IFunctionEndpoint endpoint)
  {
      this.endpoint = endpoint;
  }

  private const AutoComplete = false;

  [FunctionName("MyEndpointTrigger")]
  public async Task Run(
        [ServiceBusTrigger(queueName: "MyEndpoint", AutoComplete: AutoComplete)]
        Message message,
        MessageReceiver messageReceiver,
        ILogger logger,
        ExecutionContext executionContext)
  {
    await endpoint.Process(message, executionContext, 
              messageReceiver, autoComplete:AutoComplete, // new mandatory parameters
              logger);
    }
}
```

With this approach
1. Users don't have to choose between two methods, `Process` and `ProcessTransactional`
2. The signature method is consistent and always passes in `MessageSender` and the value for the auto-complete is used as a transport transaction mode indicator\*

\*The value of the flag would be validated against the value provided by the `NServiceBusEndpointEnableCrossEntityTransactions` attribute when users choose to override the trigger functions and do not use `NServiceBusEndpointName` assembly attribute.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
```diff
@@ -28,14 +30,97 @@ public class FunctionEndpoint : IFunctionEndpoint
...
+            {
+                try
+                {
+                    using (var transaction = new CommittableTransaction(new TransactionOptions { IsolationLevel = IsolationLevel.Serializable, Timeout = TransactionManager.MaximumTimeout }))
+                    {
+                        TransportTransaction transportTransaction = CreateTransportTransaction(message, messageReceiver, transaction);
+
+                        MessageContext messageContext = CreateMessageContext(message, transportTransaction);
+
+                        //TODO: Could also be done earlier since the token here doesn't need to come from the context (it's a new one anyway)
```
mikeminutillo, line 50: This only matters if the initialize takes significant time. If it does, we coud start it immediately and then await the result here. I'm guessing this may change in v8 when we can pass a cancellation token in.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
```diff
@@ -65,19 +150,17 @@ public async Task Process(Message message, ExecutionContext executionContext, IL
...
 
+        static MessageContext CreateMessageContext(Message originalMessage, TransportTransaction transportTransaction) =>
+            new MessageContext(
+                originalMessage.GetMessageId(),
+                originalMessage.GetHeaders(),
+                originalMessage.Body,
+                transportTransaction,
+                new CancellationTokenSource(),
+                new ContextBag());
+
```
mikeminutillo, line 163: If the previous PR we dead-letter the message if it fails loading the body, message Id, or headers. Is there a reason we don't do that any more?

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/IFunctionEndpoint.cs
mikeminutillo: I don't mind having two methods. It makes it more explicit whether message sending will be atomic with receives.

If we do have two methods, I'm not sure about using the term `Transactional` because as far as I am aware the original `Process` method is still doing a transactional receive. i.e. If things fail or the functions host is restarted, the message will re-appear in the queue. `EnableCrossEntityTransactions` sounds promising.

I'm not sure about the term `AutoComplete` either but I see in @SeanFeldman example that it is used in the `ServiceBusTrigger` attribute so I am guessing that it is a well-known Azure Functions term that will be understood by users?

The only thing I don't like about adding `MessageReceiver` (or a boolean flag) as a mandatory parameter is it will require a breaking change, and every user will have to update every trigger, even if they don't care about atomic sends with receives (unless they're using the attributes). We could believe this is good thing as it will force users to examine each use-case and make an explicit decision.

mikeminutillo: We could have an overload of `Process()` that takes both new parameters. We could also find the `ServiceBusTriggerAttribute` in the stack frame and verify that the `AutoComplete` property on it matches what was passed to `Process`. If we're going to do that, you could drop the requirement to pass it in and simply read the value. I don't know how computationally expensive it is to do that reflection each time.

SeanFeldman: I like the idea of having an overload of `Process()` and looking up the `ServiceBusTrigger` attribute!
@mikeminutillo, you've earned your cookie for today 😀

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
timbussmann: I think there is currently no actual reason to not start before opening the transaction as it doesn't need one. It currently uses the cancellation token generated on the context but it would probably be more likely that we want to use the CTS from the function parameters instead. The only thing to consider is an error scenario which has to ensure we abandon the message.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
timbussmann: yeah agreed.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
timbussmann: left that out for simplicty for now as we also don't do that on the current stable release IIRC.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
timbussmann: btw. an interesting difference to the existing `Process` method is that endpoint startup failures would currently cause the error pipeline to be invoked (which might also fail because the endpoint didn't start in the first place). So in cases where the endpoint doesn't start, the only "reliable" action would be to just abandon the message and doing "native" immediate retries until the message eventually gets dead-lettered

src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
```diff
@@ -0,0 +1,142 @@
...
+                    {
+                        // 2nd parameter was triggerFunctionName
+                        triggerFunctionName = AttributeParameterAtPosition(1);
+
+                        // 3rd parameter might be cross entity transaction flag
+                        enableCrossEntityTransactions = attributeParametersCount == 3 && bool.Parse(AttributeParameterAtPosition(2));
+                    }
+                }
+
+                bool IsNServiceBusEndpointNameAttribute(string value) => value.Equals("NServiceBus.NServiceBusTriggerFunctionAttribute");
```
mikeminutillo, line 74: Looking at the code above, it looks like we might pass `null` here and get a `NullReferenceException`. I'm not sure how you could get an `AttributeSyntaxNode` without a `Type` so it's probably not a big deal.

src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
```diff
@@ -0,0 +1,142 @@
...
+                        return;
+                    }
+
+                    if (bool.TryParse(AttributeParameterAtPosition(1), out enableCrossEntityTransactions))
+                    {
+                        // 2nd parameter was cross entity transaction flag
+                        // 3rd parameter might be trigger function name
+                        triggerFunctionName = attributeParametersCount == 3
+                            ? AttributeParameterAtPosition(2)
+                            : $"NServiceBusFunctionEndpointTrigger-{endpointName}";
```
mikeminutillo, line 62: Should this be extracted as a helper method? It gets referenced above too.

src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
```diff
@@ -0,0 +1,142 @@
...
+        MessageReceiver messageReceiver,
+        ILogger logger,
+        ExecutionContext executionContext)
+    {{
+        {(syntaxReceiver.enableCrossEntityTransactions
+            ? "await endpoint.ProcessTransactional(message, executionContext, messageReceiver, logger);"
+            : "await endpoint.ProcessNonTransactional(message, executionContext, messageReceiver, logger);")}
+    }}
+}}";
+            context.AddSource("NServiceBus__FunctionEndpointTrigger", SourceText.From(source, Encoding.UTF8));
```
mikeminutillo, line 139: Is it OK to use the same hintname for both generators?

src/ServiceBus.Tests/MessageHelper.cs
```diff
@@ -0,0 +1,43 @@
...
+            var fieldInfo = typeof(Message.SystemPropertiesCollection).GetField("sequenceNumber", BindingFlags.NonPublic | BindingFlags.Instance);
+            fieldInfo.SetValue(systemProperties, 123);
+            // set delivery count to 1
+            var deliveryCountProperty = typeof(Message.SystemPropertiesCollection).GetProperty("DeliveryCount");
+            deliveryCountProperty.SetValue(systemProperties, 1);
+            // assign test message mocked system properties
+            var property = typeof(Message).GetProperty("SystemProperties");
+            property.SetValue(asbMessage, systemProperties);
+
+            asbMessage.MessageId = Guid.NewGuid().ToString("N");
```
mikeminutillo, line 39: I don't think every message is guaranteed to have a `MessageId`. I noticed some code failures in one of the branches because we were missing this, and I think we have to be able to cope without it for native integrations.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/NServiceBusTriggerFunctionAttribute.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <summary>
+        /// Enable cross-entity transactions.
+        /// </summary>
+        public bool EnableCrossEntityTransactions { get; set; }
+
+        /// <summary>
+        /// Endpoint logical name.
+        /// </summary>
+        /// <param name="name">Endpoint name that is the input queue name.</param>
+        public NServiceBusTriggerFunctionAttribute(string name)
```
mikeminutillo, line 29: Should this be `endpointName` to be more specific?

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/NServiceBusTriggerFunctionAttribute.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace NServiceBus
+{
+    ///<summary>
+    /// Assembly attribute to specify NServiceBus logical endpoint name.
```
mikeminutillo, line 4:
```suggestion
    /// Assembly attribute to configure generated NServiceBus Azure Function.
```

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
```diff
@@ -28,54 +30,118 @@ public class FunctionEndpoint : IFunctionEndpoint
             endpointFactory = _ => externallyManagedContainerEndpoint.Start(serviceProvider);
         }
 
+        /// <summary>
+        /// Processes a message received from an AzureServiceBus trigger using the NServiceBus message pipeline. This method will lookup the <see cref="ServiceBusTriggerAttribute.AutoComplete"/> setting to determine whether to use transactional or non-transactional processing.
+        /// </summary>
+        public Task Process(Message message, ExecutionContext executionContext, IMessageReceiver messageReceiver, ILogger functionsLogger = null) =>
```
mikeminutillo, line 36: Should we make this an explicit interface implementation? Does it make sense to call this method if you have a direct reference to a `FunctionEndpoint`? I'm not sure that we can do the same for the deprecated version as that is a breaking change.

timbussmann: yeah we can make this an explicit interface implementation 👍

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/NServiceBusTriggerFunctionAttribute.cs
timbussmann: sounds reasonable 👍

src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
timbussmann: 🤷

src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
timbussmann: it doesn't look like we need to recompute the value here at all and can just use the initial value instead.

src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
timbussmann: I think this is not possible given we check the type of the attribute before so it's probably safe to expect the attribute list to always be present here.

src/ServiceBus.Tests/MessageHelper.cs
timbussmann: this is currently needed for some of the recoverability tests to ensure we can identify the message across retries as by default we generate new IDs each time a message without an ID is received ( see the implementation of `GetMessageId`)

HEskandari (approved):

src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
timbussmann: fixed, not sure why R# also didn't point this out tbh

PR #239, opened 2021-07-29
Capture NServiceBus endpoint startup logs before logger is available

### LINKED ISSUE #238: Logs about license file and probed locations from NServiceBus not logged with Functions
NServiceBus scans various locations for a valid license and outputs the location probed and the one selected.
All traversed locations and the scan result are logged (see https://docs.particular.net/nservicebus/licensing/#troubleshooting).
The instructions for troubleshooting are not working for the Functions package.

The issue stems from how the NServiceBus logger is adopted to Microsoft `ILogger`. This seems to be handled properly with `NServiceBus.Extensions.Hosting` by using a deferred logger factory that flushes all NServiceBus logs when an `ILogger` is available. A similar approach should be taken with the Functions package to allow license troubleshooting.

mikeminutillo:
Closes #238 

Uses the same technique as [NServiceBus.Extensions.Hosting](https://github.com/Particular/NServiceBus.Extensions.Hosting) to capture all log messages in deferred loggers until an `ILogger` instance is available.

NOTE: This will only provide startup logs when the first trigger is called.


SeanFeldman (approved):

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
```diff
@@ -34,6 +34,7 @@ public class FunctionEndpoint : IFunctionEndpoint
         public async Task Process(Message message, ExecutionContext executionContext, ILogger functionsLogger = null)
         {
             FunctionsLoggerFactory.Instance.SetCurrentLogger(functionsLogger);
```
timbussmann, line 36: we'd have to change the logger on the `LogManager` here back to the Functions logger, otherwise the Deferred Factory remains the default logger I think

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
```diff
@@ -34,6 +34,7 @@ public class FunctionEndpoint : IFunctionEndpoint
...
+            DeferredLoggerFactory.Instance.FlushAll(FunctionsLoggerFactory.Instance);
```
timbussmann, line 37: with this approach, I'm not really sure why need the whole AsyncLocal stuff in the Functions logger anymore. Tbh it feels like we could combine those two loggers into one, but I'm also ok to keep them separate for now to keep them simpler.

timbussmann (changes requested):

HEskandari (approved):

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
mikeminutillo: I think it's still possible that the same Azure Function can be triggered multiple times in parallel within the same instance. I don't know if the internal logging mechanism keeps those separate. If it does then the async local is still required.

I did look at combining these but it gets complicated as the MS Extensions `ILogger` uses generics for it's log state, which makes it harder to store them in a deferred way.

mikeminutillo: Just typing that made me realize I can defer it in NSB format and flush it in MS Extensions format. I will take another pass.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
mikeminutillo: You are right. I used to have something in the `FunctionEndpoint` factory method that did this swap but I dropped it because we can't guarantee that's only way it gets constructed. I missed putting it back in here.

mikeminutillo: Actually, it _is_ the only way to get a `FunctionEndpoint`. It's constructor is internal. :)
That simplifies things

mikeminutillo:
Collapsed back into a single logger that can defer until it is handed a logger via a call to `Process()`. Note that if you don't provide a logger then we put a `NullLogger` in place. This prevents deferred logs from building up indefinitely.

There _is_ still a case where logs could grow indefinitely. That's where you're using `FunctionEndpoint` but never calling `Process()`.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
```diff
@@ -33,7 +34,7 @@ public class FunctionEndpoint : IFunctionEndpoint
         /// </summary>
         public async Task Process(Message message, ExecutionContext executionContext, ILogger functionsLogger = null)
         {
-            FunctionsLoggerFactory.Instance.SetCurrentLogger(functionsLogger);
+            FunctionsLoggerFactory.Instance.SetCurrentLogger(functionsLogger ?? NullLogger.Instance);
```
timbussmann, line 37: could we do that fallback inside `SetCurrentLogger`?

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/Logging/Logger.cs
```diff
@@ -14,89 +14,99 @@ public Logger(AsyncLocal<ILogger> logger)
...
+        void Log(LogLevel level, string message)
         {
-            CurrentLogger.Log(LogLevel.Debug, message);
+            var concreteLogger = logger.Value;
+            if (concreteLogger == null)
+            {
+                deferredMessageLogs.Enqueue((level, message));
+                return;
+            }
+            Flush(concreteLogger);
```
timbussmann, line 31: what's the scenario that requires Flush to be called here given that Flush is also called when setting the logger in `SetCurrentLogger`?

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
mikeminutillo: Done

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/Logging/Logger.cs
mikeminutillo: I can't think of one. The flush in `SetCurrentLogger` was added during testing.

mikeminutillo:
Smoke testing and this appears to work 

image

This also highlights the fact that all of the sending/publishing APIs can also be passed a logger and this works as expected.

HEskandari (approved):

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/Logging/Logger.cs
```diff
@@ -14,89 +14,96 @@ public Logger(AsyncLocal<ILogger> logger)
...
-        public bool IsFatalEnabled => CurrentLogger.IsEnabled(LogLevel.Critical);
-
-        public void Debug(string message)
+        void Log(LogLevel level, string message)
         {
-            CurrentLogger.Log(LogLevel.Debug, message);
+            var concreteLogger = logger.Value;
+            if (concreteLogger == null)
+            {
+                deferredMessageLogs.Enqueue((level, message));
```
timbussmann, line 28: is there a potential race condition in a case where a caller entered the if condition and then concurrently the logger is set and the queues are emptied before the caller enters the enqueue method, essentially causing a "log loss"?

timbussmann (approved):
one thought but otherwise LGTM

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/Logging/Logger.cs
mikeminutillo: The concrete logger comes from an async local. If another thread sets it then it will be in that threads async local context so it should not affect this one.

timbussmann: hm not sure that's what I mean, here's what I was thinking about:

Thread1(T1): checks `concreteLogger`, value is `null`, enters if condition.
Thread2(T2): calls `SetCurrentLogger`
T2: Sets logger value to logger instance
T2: Subsequentlly calls `Flush`
T2: Flush empties all `defferedMessageLogs` entries
T2: completes
T1: adds log entry to `defferedMessageLogs` as part of the previously entered if-block. `defferedMessageLogs` isn't read anymore.

mikeminutillo: We used to flush on every log call which would resolve this problem. I can put that back. Given that the deferred logger will usually be empty, this shoud be quick. We'd create three enumerators, for every log call which we can probably avoid if we check the queue length before enumerating them.

mikeminutillo: Actually, because it's a queue we only ever `TryDequeue`. There's no enumerator created. This should be quick.

timbussmann: not really sure that's worthwhile though. Arguably, there shouldn't really be log statements being created while the logger is being replaced as we haven't started processing messages yet and the endpoint started up already, so maybe this is something that doesn't necessarily need to be fixed. The only potential log statements would come from sort of background job which shouldn't really happen without an active message pump.

timbussmann:
merging this for now to start testing. @mikeminutillo if you want to follow-up on our discussion regarding the race condition, we can also do that via a separate PR I think

PR #240, opened 2021-07-29
Generate trigger when using transactions

SeanFeldman:
Add support for cross-entity transactions configuration via a new `NServiceBusTriggerFunctionAttribute` attribute.

TODO: 
- [x] obsolete `NServiceBusEndpointNameAttribute` with warning.


src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
```diff
@@ -0,0 +1,139 @@
...
+
+    [FunctionName(""{syntaxReceiver.triggerFunctionName}"")]
+    public async Task Run(
+        [ServiceBusTrigger(queueName: ""{syntaxReceiver.endpointName}"", AutoComplete: {(!syntaxReceiver.enableCrossEntityTransactions).ToString().ToLowerInvariant()})]
+        Message message,
+        MessageReceiver messageReceiver,
+        ILogger logger,
+        ExecutionContext executionContext)
+    {{
+        await endpoint.ProcessTransactional(message, executionContext, messageReceiver, logger);
```
timbussmann, line 133: shouldn't this switch between the Process and ProcessTransactional method depending on the transaction configuration?

src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
```diff
@@ -0,0 +1,139 @@
...
+                    enableCrossEntityTransactions = false;
+
+                    var attributeParametersCount = AttributeParametersCount();
+
+                    if (attributeParametersCount == 1)
+                    {
+                        return;
+                    }
+
+                    if (bool.TryParse(AttributeParameterAtPosition(1), out enableCrossEntityTransactions))
```
timbussmann, line 56: this is based on the assumption that we will not introduce another bool parameter in the attribute right? If there would be multiple bool values, we wouldn't be able to distinguish them furher.

timbussmann:
the parameter and the property itself allowed the same config setting to be configured in two ways, e.g.  `[Attribute("mandatory", "optional ctor param")]` or `[Attribute("mandatory", SomeProperty = "property value")]`. The second is support for attributes ootb. Given the second one is a bit more expressive, I've went with only supporting properties to remove confusion and ambiguity.

timbussmann (approved):

src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
SeanFeldman: Let's cross that bridge when get there 🙂

src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
```diff
@@ -0,0 +1,142 @@
...
+
+    [FunctionName(""{syntaxReceiver.triggerFunctionName}"")]
+    public async Task Run(
+        [ServiceBusTrigger(queueName: ""{syntaxReceiver.endpointName}"", AutoComplete = {(!syntaxReceiver.enableCrossEntityTransactions).ToString().ToLowerInvariant()})]
+        Message message,
+        MessageReceiver messageReceiver,
+        ILogger logger,
+        ExecutionContext executionContext)
+    {{
+        {(syntaxReceiver.enableCrossEntityTransactions
```
SeanFeldman, line 134: This is why I'm in favour of using a single method all the time and pass `messageReceiver` no matter what.
1. Less logic here
1. Less cognitive load for customers that choose to override the trigger - no need to distinguish between modes and always a single method/same parameters to pass in.

src/NServiceBus.AzureFunctions.SourceGenerator.Tests/SourceGeneratorApprovals.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace NServiceBus.AzureFunctions.SourceGenerator.Tests
+﻿#pragma warning disable IDE0079 // Remove unnecessary suppression
```
SeanFeldman, line 1: Is it possible to do these pragmas with `.editorconfig`?

SeanFeldman: Ah, nm, it's obsoleted. 👍

SeanFeldman:
I'll go ahead and merge but suggest looking into the [comment](https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/pull/240/files#r679182438) as I feel we're inflicting unnecessary work on the segment of users that want to override the trigger.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/NServiceBusTriggerFunctionAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <summary>
+        /// Override trigger function name.
+        /// </summary>
+        public string TriggerFunctionName { get; set; }
+
+        /// <summary>
+        /// Enable cross-entity transactions.
+        /// </summary>
+        public bool EnableCrossEntityTransactions { get; set; }
```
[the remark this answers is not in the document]
timbussmann: > Not to mention that EnableTransactions would mislead users to believe they are fully transactional.

agreed

> However, I would not want to call it anything like the transport transaction mode as that term is really specific to NServiceBus

that might be good though, assuming they are familiar with NSB, it would be fairly clear to them what this means.

>  I wouldn't even mind calling it AutoComplete

I think I wouldn't do that given AutoComplete isn't directly tied to transactions, just something that is a "side effect" of using transactions.

src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
timbussmann: shoutout to our future selves 😆

src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
timbussmann: happy to discuss that further, I don't have strong opinions on this.

> Less cognitive load for customers that choose to override the trigger 

that might be dangerous for customers that write their own trigger as long as we don't have better support to detect whether `AutoComplete` has been properly configured.

SeanFeldman: Using a canonical signature to pass in `MessageReceiver`, how is that dangerous?

PR #241, opened 2021-08-02
Use single process implementation

timbussmann:
Refactoring attmept to use a single "Process" implementation by both transactional and non-transactional methods. Potentially opening up for more public `Process` overloads that can determine transactional/non-transactional behavior at runtime.

Not sure I really like the generalization, the amount of callbacks to handle the tx handling is quite ugly over the benefit of a single code path.

Thoughts?

### DIFF
--- a/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
+++ b/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
@@ -94,6 +94,8 @@
             await InitializeEndpointIfNecessary(functionExecutionContext, CancellationToken.None)
                 .ConfigureAwait(false);
 
+            var messageId = message.GetMessageId();
+
             try
             {
                 using (var transaction = transactionFactory())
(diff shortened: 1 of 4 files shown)


SeanFeldman (commented):
Personally, I'm ok with this change because it reduces the complexity for the end users. But if you feel like taking a shower after this PR, another option is keep a single method for the public API and separate paths internally.

timbussmann:
> nother option is keep a single method for the public API and separate paths internally.

that is a change that I think is completely decoupled from each other. Both public API and internal implementation can be handled completely independently in regards to code reuse and naming.

HEskandari (approved):
Approved. What I noted is that this isn't covered by tests, is that correct? Can this be tested?

mikeminutillo (approved):
I don't mind this. I think it makes the Process method very readable. We could pull the callbacks out into a `FunctionsTransactionStrategy` or something which could simplify things inside of `FunctionEndpoint`. That would further separate the implementation though which may make it harder to follow. I pushed a PR showing what that looks like #243

timbussmann:
> Approved. What I noted is that this isn't covered by tests, is that correct? Can this be tested?

there are the existing acceptance tests but those do not test any transactional behavior. I'm currently looking into this but will probably scoped to a dedicated PR.

timbussmann:
made some smaller changes that allow at least partially testing some of the process logic. Some stuff is still hard to test but maybe @mikeminutillo spike #243 might help with that. I'll merge this into the main PR for now based on the feedback

PR #242, opened 2021-08-02
Add transaction tests

timbussmann:
WIP, trying to create a dedicated test project to avoid having to use `TestCase` or duplicate tests everywhere


timbussmann:
Closing this as this won't fly given that the outgoing messages are using the actual transport while the incoming trigger is invoked manually without the message actually being on the broker. Therefore transactional tests won't really work anyway.

This would not only require a real message to be sent to ASB to start with, but also the actual trigger infrastructure to read and invoke the function logic. 

Closing this for now and trying to find alternative approaches to test this.

PR #244, opened 2021-08-03
Auto-detect AutoComplete setting

timbussmann:
Adding a new Process method to the `IFunctionEndpoint` interface while obsoleting the old process method. The new method which is intended to be called by users that do not use the auto-generated trigger will detect the `AutoComplete` setting by walking up the CallStack.

This comes at an additional computation & allocation cost but prevents users from using the wrong tx/non-tx method with incompatible AutoComplete settings.
The `FunctionEndpoint` class itself exposes the explicit operations to do transactional/non-transactional processing for advanced uses cases. This is also the API used by the generated triggers.


timbussmann:
a rough test with Benchmark.net indicates a ~30ms computation overhead and ~7KB extra allocations caused by the stackwalk logic per invocation.

mikeminutillo:
What if we put a version with a boolean parameter behind an `.Advanced` property. So you can either do:

`functionEndpoint.Advanced.Process(message, executionContext, messageReceiver, autoComplete, logger)` and we'll obey the `autoComplete` setting provided by the caller.

Or:

`functionEndpoint.Process(message, executionContext, messageReceiver, logger)` and we'll try to infer `autoComplete` from the attribute and use it to call the advanced version.

If we cannot find the attribute, the exception message can point the user to the advanced method. Our generated code can jump straight to using the advanced method. People who do not want the overhead caused by the trigger-finding code can fall back to using the advanced method. Users who are hand-generating the function are unlikely to call the advanced method.

I spiked what this would look like in #247

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
```diff
@@ -78,6 +88,15 @@ TransportTransaction CreateTransportTransaction(CommittableTransaction transacti
         /// <summary>
         /// Processes a message received from an AzureServiceBus trigger using the NServiceBus message pipeline.
         /// </summary>
+        public Task ProcessNonTransactional(Message message, ExecutionContext executionContext, IMessageReceiver messageReceiver, ILogger functionsLogger = null) => Process(message, executionContext, functionsLogger);
```
mikeminutillo, line 91: Should we be calling in to an obsolete method? We're going to have to fix it as soon as we update to 2.0 anyway.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ReflectionHelper.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    using System;
+    using System.Diagnostics;
+    using System.Reflection;
+    using Microsoft.Azure.ServiceBus;
+    using Microsoft.Azure.WebJobs;
+
+    class ReflectionHelper
+    {
+        public static bool GetAutoCompleteValue()
```
mikeminutillo, line 11: There already exists a `TriggerDiscoverer` class in the project that does something very similar to this (it walks the stack trace frame looking for `ServiceBusTriggerAttribute`). Is it worth re-using that here? I think it's only used by `.FromAttributes()` which is going away.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ReflectionHelper.cs
```diff
@@ -0,0 +1,35 @@
...
+            var frames = st.GetFrames();
+            foreach (var frame in frames)
+            {
+                var method = frame?.GetMethod();
+                if (method?.GetCustomAttribute<FunctionNameAttribute>() != null)
+                {
+                    foreach (var parameter in method.GetParameters())
+                    {
+                        ServiceBusTriggerAttribute serviceBusTriggerAttribute;
+                        if (parameter.ParameterType == typeof(Message)
```
mikeminutillo, line 23: Is it possible to put a ServiceBusTriggerAttribute on any other type of parameter? What about when switching to the new SDK

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ReflectionHelper.cs
```diff
@@ -0,0 +1,35 @@
...
+                            && (serviceBusTriggerAttribute = parameter.GetCustomAttribute<ServiceBusTriggerAttribute>()) != null)
+                        {
+                            return serviceBusTriggerAttribute.AutoComplete;
+                        }
+                    }
+                }
+            }
+
+            throw new Exception($"Could not locate {nameof(ServiceBusTriggerAttribute)} to infer AutoComplete setting.");
```
mikeminutillo, line 32: Should we recommend the advanced approach here?

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ReflectionHelper.cs
SeanFeldman: > Is it possible to put a ServiceBusTriggerAttribute on any other type of parameter?

No.

> What about when switching to the new SDK

Same deal

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
timbussmann: it was just to not duplicate the code because the obsoleted message still needs to work and it can't call `ProcessNonTransactional` because of the missing receiver (or pass null but I didn't like that). I'm fine either way.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ReflectionHelper.cs
timbussmann: good point, will have  a look

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ReflectionHelper.cs
timbussmann: > ServiceBusTriggerAttribute on any other type of parameter

yes. e.g. the samples on the official docs use a string input parameter instead: https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-service-bus-trigger?tabs=csharp#example

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ReflectionHelper.cs
timbussmann: makes sense 👍

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ReflectionHelper.cs
SeanFeldman: A parameter is always a parameter for the message. The type of the parameter can be `byte[]`, `string`, or a POCO to deserialize to. With the in-proc SDK, it can also be `Message` (ASB SDK). With isolated worker SDK, SDK type cannot be used. Keep that in mind.

timbussmann: yes but given our API needs we have a strong need to enforce this to use the ASB SDK message here. For the isolated worker, well have to come up with something different anyway I guess.

SeanFeldman: > yes but given our API needs we have a strong need to enforce this to use the ASB SDK message here. For the isolated worker, well have to come up with something different anyway I guess.

We do not need to use the ASB SDK message.

HEskandari (approved):

mikeminutillo (approved):
Happy for this to go to the transactions branch

PR #247, opened 2021-08-05
[Spike] Auto detect with Advanced API

mikeminutillo:
A proposal for #244

### DIFF
--- a/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/IFunctionEndpoint.cs
+++ b/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/IFunctionEndpoint.cs
@@ -14,6 +14,11 @@
     public interface IFunctionEndpoint
     {
         /// <summary>
+        /// Access advance message processing methods.
+        /// </summary>
+        IAdvancedFunctionEndpoint Advanced { get; }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


src/NServiceBus.AzureFunctions.InProcess.ServiceBus/IFunctionEndpoint.cs
```diff
@@ -13,6 +13,11 @@
     /// </summary>
     public interface IFunctionEndpoint
     {
+        /// <summary>
+        /// Access advance message processing methods.
+        /// </summary>
+        IAdvancedFunctionEndpoint Advanced { get; }
```
SeanFeldman, line 19: Clever but is this really necessary? `ServiceBusTriggeredEndpointConfiguration` has `AdvancedConfiguration` now another `Advanced` on `IFunctionEndpoint`? I really would not want to go this route.

mikeminutillo: The intent was just to put some "syntactic vinegar" on it to stop users from using the overload unless they know what they're doing with it. We discussed it and I don't think we're going with this approach.

SeanFeldman: > We discussed it and I don't think we're going with this approach.

👍

PR #256, opened 2021-08-11
Upgrade to NServiceBus version 8

### LINKED ISSUE #245: Azure Functions master is not on NServiceBus v8
## Relevant information

Azure Functions has been GA-ed recently and as a Preview package was locked with the Core version it was created with - NServiceBus version 7. Now that it's GA-ed (`NServiceBus.AzureFunctions.InProcess.ServiceBus` 1.0 is out), the `master` branch should be targeting Core v8.

Note that the `master` branch already had that change and it [had to be reverted](https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/pull/178) to allow the 1.0 release. The [PR](https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/pull/178) could be reverted to move the `master` back to Core v8.

In addition to that, the `Particular.Analyzers` package that will be in the `master` will need to be updated to remove the lock on a specific version. This is due to the GA taking place before `CancellationToken`-* TF was done and `master` branch being on the Core v7. To accomplish that, the [`Custom.Build.props`](https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/pull/185/files#diff-74df371687a523dfcb3bfdb13bad4803a7dee39772f9257d6a3d009e4196699c) containing the locked `Particular.Analyzers` version would need to be adjusted or removed to no longer lock on a specific version.

mikeminutillo:
Closes #245


timbussmann:
haven't had the chance to really look at the changes, I just wanted to ask whether you had a look at the original v8 migration that was done a while ago via https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/pull/128?

mikeminutillo:
@timbussmann I have made some changes based on your review comments. Please take another look.

I was thinking that we should consider exposing `UseTransport()` and then we can remove the `Transport` property. If you provide a connection string using one of the standard mechanisms then we can set the transport up for you and expose the routing settings on a property.

If you don't provide a connection string in the standard way, then you can call `UseTransport(AzureServiceBusTransport)` and configure it however you want.

If you don't provide it either way then core will blow up and tell you need to configure a transport.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/Serverless/TransportWrapper/PipelineInvoker.cs
```diff
@@ -1,45 +1,37 @@
...
 
-        void IPushMessages.Start(PushRuntimeSettings limitations)
-        {
-        }
+        public Task<ErrorHandleResult> PushFailedMessage(ErrorContext errorContext) => onError(errorContext);
+
+        public Task PushMessage(MessageContext messageContext) => onMessage.Invoke(messageContext);
 
-        Task IPushMessages.Stop()
+        public Task Initialize(PushRuntimeSettings limitations, OnMessage onMessage, OnError onError,
```
timbussmann, line 18: is there no need to initialize the `baseTransportReceiver` here?

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/Serverless/TransportWrapper/PipelineInvoker.cs
```diff
@@ -1,45 +1,37 @@
...
-        }
+        public Task StartReceive(CancellationToken cancellationToken = new CancellationToken()) => Task.CompletedTask;
 
-        public Task PushMessage(MessageContext messageContext)
-        {
-            return onMessage.Invoke(messageContext);
-        }
+        public Task StopReceive(CancellationToken cancellationToken = new CancellationToken()) => Task.CompletedTask;
+
+        public ISubscriptionManager Subscriptions => baseTransportReceiver.Subscriptions;
```
timbussmann, line 30: this connects to the initialize question above. It seems the subscription manager itself is created as part of the ctor and is accessible but it `Initialize` does some things with the underlying subscription manager.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/Serverless/TransportWrapper/ServerlessTransport.cs
```diff
@@ -1,33 +1,46 @@
...
+            PipelineInvoker = (PipelineInvoker)serverlessTransportInfrastructure.Receivers[MainReceiverId];
+            return serverlessTransportInfrastructure;
         }
 
-        readonly TBaseTransport baseTransport;
+        public PipelineInvoker PipelineInvoker { get; private set; }
+
+        public override string ToTransportAddress(QueueAddress address) => baseTransport.ToTransportAddress(address);
+
+        public override IReadOnlyCollection<TransportTransactionMode> GetSupportedTransactionModes() => baseTransport.GetSupportedTransactionModes();
```
timbussmann, line 42: should we implement this to only return `ReceiveOnly` and `SendsAtomicWithReceive`?

src/NServiceBus.AzureFunctions.SourceGenerator.Tests/SourceGeneratorApprovals.cs
```diff
@@ -1,173 +0,0 @@
-﻿#pragma warning disable IDE0079 // Remove unnecessary suppression
```
timbussmann, line 1: have you also deleted the related approved txt files?

timbussmann:
> I was thinking that we should consider exposing UseTransport() and then we can remove the Transport property. If you provide a connection string using one of the standard mechanisms then we can set the transport up for you and expose the routing settings on a property.

> If you don't provide a connection string in the standard way, then you can call UseTransport(AzureServiceBusTransport) and configure it however you want.

wouldn't that mean that every transport configuration via the `Transport` property now would require the user to setup the transport  (and fetching the connection string) manually? I think having a simple access to configure some common transport settings without having to deal with the connection string would make sense.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/Serverless/TransportWrapper/ServerlessTransport.cs
SeanFeldman: Yep. `ReceiveAndDelete` is not supported so it would not work anyways.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/Serverless/TransportWrapper/PipelineInvoker.cs
mikeminutillo: Added. Followed the same pattern as the prior PR by passing in our own no-ops. We could probably just pass the real `OnMessage` and `OnError` because they will never be called. That way if the signature changes we don't need to make a change here.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/Serverless/TransportWrapper/ServerlessTransport.cs
mikeminutillo: Done

src/NServiceBus.AzureFunctions.SourceGenerator.Tests/SourceGeneratorApprovals.cs
mikeminutillo: Have now :)

timbussmann (approved):

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
```diff
@@ -76,12 +75,11 @@ internal ServiceBusTriggeredEndpointConfiguration(string endpointName, string co
                 EndpointConfiguration.License(licenseText);
             }
 
-            Transport = UseTransport<AzureServiceBusTransport>();
-
             var connectionString = GetConfiguredValueOrFallback(configuration, connectionStringName ?? DefaultServiceBusConnectionName, optional: true);
             if (!string.IsNullOrWhiteSpace(connectionString))
             {
-                Transport.ConnectionString(connectionString);
+                // TODO: What do we do if we don't have a connectionString?
```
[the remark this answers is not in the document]
mikeminutillo: Now throws

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
```diff
@@ -117,7 +115,7 @@ static string GetConfiguredValueOrFallback(IConfiguration configuration, string
         /// <summary>
         /// Azure Service Bus transport
         /// </summary>
-        public TransportExtensions<AzureServiceBusTransport> Transport { get; }
+        public AzureServiceBusTransport Transport { get; }
```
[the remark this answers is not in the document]
mikeminutillo: We ended up getting around this by replacing the property with a method that accepts a Transport configuration action.

```chsarp
serviceBusTriggeredEndpointConfiguration.ConfigureTransport(transport =>...);
```

There's also a routing version

```csharp
serviceBusTriggeredEndpointConfiguration.Routing(routing => ...);
```

Issue #22, opened 2019-08-29
Add CancellationToken support to the Process method


SeanFeldman:
Azure Functions [support `CancellationToken`](https://docs.microsoft.com/en-us/azure/azure-functions/functions-dotnet-class-library#cancellation-tokens) that should be accepted by NServiceBus.AzureFunctions.

Has dependency on NServiceBus Core [support for `CancellationToken`s](https://github.com/Particular/NServiceBus/issues/3785).

timbussmann:
our pipeline doesn't really support cancellation tokens at the moment, so there isn't much to gain by passing this in besides maybe aborting the startup behavior. Not sure that's worth it for the noise it creates (and the expectations from the user).

SeanFeldman:
This issue is here as a placeholder until Core has support for the cancellation token. It's not intended for public preview (Core v7).

timbussmann:
I'm not sure it's desirable to keep placeholder issues open for features which aren't even scheduled at this point. We can always open this once this gets into the foreseeable timeline?

PR #272, opened 2021-08-18
Support cancellation token

HEskandari:
Resolves https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/issues/22

Add support for cancellation token to the process method.


src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
```diff
@@ -71,9 +73,8 @@ public async Task ProcessNonTransactional(Message message, ExecutionContext exec
             await InitializeEndpointIfNecessary(functionExecutionContext, CancellationToken.None)
```
mikeminutillo, line 73: Should we pass the cancellation token here?

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/IFunctionEndpoint.cs
```diff
@@ -16,7 +18,7 @@ public interface IFunctionEndpoint
         /// <summary>
         /// Processes a message received from an AzureServiceBus trigger using the NServiceBus message pipeline. This method will lookup the <see cref="ServiceBusTriggerAttribute.AutoComplete"/> setting to determine whether to use transactional or non-transactional processing.
         /// </summary>
-        Task Process(Message message, ExecutionContext executionContext, IMessageReceiver messageReceiver, ILogger functionsLogger = null);
+        Task Process(Message message, ExecutionContext executionContext, IMessageReceiver messageReceiver, ILogger functionsLogger = null, CancellationToken cancellationToken = default);
```
mikeminutillo, line 21: Technically this is a breaking change. This is a major release so I'm not too concerned about that but we should agree that's what we want

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
HEskandari: done

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/IFunctionEndpoint.cs
HEskandari: Agreed.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/Serverless/TransportWrapper/PipelineInvoker.cs
```diff
@@ -11,21 +11,21 @@ public PipelineInvoker(IMessageReceiver baseTransportReceiver)
...
             this.onMessage = onMessage;
             this.onError = onError;
             return Task.CompletedTask;
         }
 
-        public Task StartReceive(CancellationToken cancellationToken = new CancellationToken()) => Task.CompletedTask;
+        public Task StartReceive(CancellationToken cancellationToken = default) => Task.CompletedTask;
 
-        public Task StopReceive(CancellationToken cancellationToken = new CancellationToken()) => Task.CompletedTask;
+        public Task StopReceive(CancellationToken cancellationToken = default) => Task.CompletedTask;
```
timbussmann, line 28: do these two even need a value to be assigned? I'd assume the interface definition already does that, and for direct calls it wouldn't matter anyway because we never call it directly anyway?

timbussmann (approved):

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/Serverless/TransportWrapper/PipelineInvoker.cs
HEskandari: No, they don't, I was going for consistency. Happy to revert it back?

timbussmann: I don't think we need to "revert back", but to completely remove the optionality of the parameter (`public Task StopReceive(CancellationToken cancellationToken)`.

HEskandari: Updated

mikeminutillo (commented):
There's a few optional cancellation tokens still that I think are only called by core or used internally. We can drop them from being optional. The only ones we need to keep are the ones on FunctionEndpoint which we are expecting the customer to call directly.

mikeminutillo (approved):

PR #288, opened 2021-08-26
Remove reflection

mikeminutillo:
Requires changes proposed in  https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/385


src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
```diff
@@ -80,8 +82,9 @@ internal ServiceBusTriggeredEndpointConfiguration(string endpointName, IConfigur
                 endpointConfiguration.License(licenseText);
             }
 
+            transport = new ServerlessAzureServiceBusTransport();
             connectionString = GetConfiguredValueOrFallback(configuration, DefaultServiceBusConnectionName, optional: true);
-            Transport = new AzureServiceBusTransport(connectionString ?? "missing"); // connection string will be checked before creating the startable endpoint as it might be configured within UseNServiceBus().
+            transport.ChangeConnectionString(connectionString);
```
timbussmann, line 87: should this be done only if connectionstring is not null/empty? There doesn't seem to be  Guard in the property setter but there might (even should?) be one?

timbussmann (approved):

timbussmann (commented):
the editorconfig file should probably be readded again

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/ServiceBusTriggeredEndpointConfiguration.cs
mikeminutillo: I put the test in place here. I did not add a guard, as there's one where the public API sets this value.

PR #332, opened 2021-10-08
Update to NServiceBus.Transports.AzureServiceBus 2.0.0

andreasohlund:


src/NServiceBus.AzureFunctions.InProcess.ServiceBus/NServiceBusEndpointNameAttribute.cs
```diff
@@ -8,39 +8,7 @@
...
-        ///<summary>
-        ///Endpoint logical name.
-        ///</summary>
-        ///<param name="name">Endpoint name that is the input queue name.</param>
-        ///<param name="triggerFunctionName">Name given to the auto-generated trigger function.</param>
-        public NServiceBusEndpointNameAttribute(string name, string triggerFunctionName)
-        {
-            Name = name;
-            TriggerFunctionName = triggerFunctionName;
-        }
```
mauroservienti, line 44: Why are these changes needed?

andreasohlund: The type was obsoleted with ERR so we usually clean out the type as much as possible since it can't be used

src/ServiceBus.Tests/When_publishing_event_from_function.cs
```diff
@@ -24,14 +24,14 @@ class Context : ScenarioContext
             public bool EventReceived { get; set; }
         }
 
-        class SubscriberEndpoint : EndpointConfigurationBuilder
+        class InsideSubscriberEndpoint : EndpointConfigurationBuilder
```
andreasohlund, line 27: @mauroservienti @WojcikMike FYI this was needed for the worker functions package as well. Since the subscriber name and the event type exists in another test as well the subscriptions gets screwed up. Not sure why this wasn't a problem with the old transport.

PR #333, opened 2021-10-10
v2.0.0 Cleanup

andreasohlund:
Realized that I had missed some cleanup of obsoleted things in v1. No need to release a 2.0.1 for this


andreasohlund:
Think this might need to be a 2.0.1 now that I think of it :(

PR #371, opened 2021-11-29
Update to latest Core 8.0.0-alpha.1900

andreasohlund:


src/ServiceBus.Tests/ServiceBus.Tests.csproj
```diff
@@ -15,12 +15,13 @@
   <ItemGroup>
     <PackageReference Include="GitHubActionsTestLogger" Version="1.2.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
+    <PackageReference Include="NServiceBus.Newtonsoft.Json" Version="3.0.0-alpha.156" />
```
timbussmann, line 18: why do we need to explicitly include the json package on all the test projects?

timbussmann (approved):

src/ServiceBus.Tests/ServiceBus.Tests.csproj
andreasohlund: I remember some old guidelines where we said to have the test reference the latest version where the main package has a version range dependency to force dependabot updates to the tests? (can't find any docs tho)

andreasohlund: Removed now

timbussmann: I do think that's true, to enable dependabot to raise update for the tests. I've seen that almost all packages that I updated didn't have any explicit references to NSB or other packages that would allow for that and here it also just contains JSON, so it would inconsistent too. (made a note on this for the retro)

src/Testing.Handlers/Testing.Handlers.csproj
```diff
@@ -5,7 +5,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="NServiceBus" Version="8.0.0-alpha.1898" />
+    <PackageReference Include="NServiceBus" Version="8.0.0-alpha.1900" />
+    <PackageReference Include="NServiceBus.Newtonsoft.Json" Version="3.0.0-alpha.156" />
```
timbussmann, line 9: again, I actually think it would be better to have explicit package references for dependabot but let's keep it consistent for now?

```suggestion
```

timbussmann (approved):

Issue #358, opened 2021-11-18
Testable FunctionEndpoint


rineshpk:
[NServiceBus.Testing](https://docs.particular.net/samples/unit-testing/) is great for unit testing handlers and endpoint. It provides a testable EndpointInstance. It would be great if there is a testable FunctionEndpoint also to assert on published or sent messages using the Function endpoint.

HEskandari:
@rineshpk Thanks for raising this. We'll have a closer look at the issue when we're running the next enhancement release in this area.

PR #378, opened 2021-12-04
Separate message processing  from interface used to send message

andreasohlund:
This separates the part users should care about from the processing part which should only be used when having to work around bugs in our source generator.

This makes https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/issues/358 easier as well since we don't have to provide a way to test the processing part.

### DIFF
--- a/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionsHostBuilderExtensions.cs
+++ b/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionsHostBuilderExtensions.cs
@@ -90,6 +90,7 @@
             functionsHostBuilder.Services.AddSingleton(serviceBusTriggeredEndpointConfiguration);
             functionsHostBuilder.Services.AddSingleton(startableEndpoint);
             functionsHostBuilder.Services.AddSingleton<IFunctionEndpoint, InProcessFunctionEndpoint>();
+            functionsHostBuilder.Services.AddSingleton<IMessageProcessor, InProcessFunctionEndpoint>();
         }
 
         internal static IStartableEndpointWithExternallyManagedContainer Configure(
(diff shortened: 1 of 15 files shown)


src/NServiceBus.AzureFunctions.InProcess.ServiceBus/obsoletes-v3.cs
```diff
@@ -13,7 +13,7 @@ public partial class ServiceBusTriggeredEndpointConfiguration
 
 namespace NServiceBus
 {
-    [ObsoleteEx(ReplacementTypeOrMember = nameof(IFunctionEndpoint),
+    [ObsoleteEx(ReplacementTypeOrMember = nameof(IMessageProcessor),
```
andreasohlund, line 16: This is the most likely replacement since users would have used the old `FunctionEndpoint` type to do processing of messages

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionsHostBuilderExtensions.cs
```diff
@@ -90,6 +90,7 @@ public static partial class FunctionsHostBuilderExtensions
             functionsHostBuilder.Services.AddSingleton(serviceBusTriggeredEndpointConfiguration);
             functionsHostBuilder.Services.AddSingleton(startableEndpoint);
             functionsHostBuilder.Services.AddSingleton<IFunctionEndpoint, InProcessFunctionEndpoint>();
+            functionsHostBuilder.Services.AddSingleton<IMessageProcessor, InProcessFunctionEndpoint>();
```
danielmarbach, line 93: This doesn't work. You would get two instances of the `InProcessFunctionEndpoint`. To achieve what you want to achieve, you would need to use the lambda overload again :( Which is as we know broken at the moment

https://andrewlock.net/how-to-register-a-service-with-multiple-interfaces-for-in-asp-net-core-di/

andreasohlund: Shit, you are right :(

andreasohlund:
Blocked until https://github.com/Azure/azure-functions-host/issues/7874 is fixed and released

andreasohlund:
I'll reopen this once that has happened

Issue #359, opened 2021-11-18
Functions Host v4 is not supported


andreasohlund:
We require the root `IServiceProvider` to be available when initializing the endpoint. This is done on the first request but Host v4 now executes functions with a scoped provider. This means that we will get access to a IServiceProvider that will be disposed of after the first call resulting in 
 
```
[2021-11-18T14:58:45.221Z] Executed 'NServiceBusFunctionEndpointTrigger-InProcess-HostV4' (Failed, Id=acd86500-4424-4506-90ea-b5aa23cf94c8, Duration=56ms)
[2021-11-18T14:58:45.223Z] System.Private.CoreLib: Exception while executing function: NServiceBusFunctionEndpointTrigger-InProcess-HostV4. Microsoft.Azure.WebJobs.Script.WebHost: Container is disposed and should not be used: Container is disposed.
```

andreasohlund:
FYI @mauroservienti @jenspettersson @danielmarbach @timbussmann

PR #381, opened 2021-12-08
Adjust DI usage to workaround Host v4 bug

andreasohlund:
Backport of fix for #359 to the release-2.1 branch.


src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
```diff
@@ -22,8 +22,13 @@
...
     {
-        // This ctor is used for the FunctionsHostBuilder scenario where the endpoint is created already during configuration time using the function host's container.
-        internal FunctionEndpoint(IStartableEndpointWithExternallyManagedContainer externallyManagedContainerEndpoint,
+        /// <summary>
+        /// This ctor is used for the FunctionsHostBuilder scenario where the endpoint is created already during configuration time using the function host's container.
+        /// </summary>
+        /// <param name="externallyManagedContainerEndpoint"></param>
+        /// <param name="configuration"></param>
+        /// <param name="serviceProvider"></param>
+        public FunctionEndpoint(IStartableEndpointWithExternallyManagedContainer externallyManagedContainerEndpoint,
```
andreasohlund, line 31: Needs to be public so that MS DI can create it

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionsHostBuilderExtensions.cs
```diff
@@ -48,29 +48,28 @@ public static class FunctionsHostBuilderExtensions
...
             IServiceCollection serviceCollection,
-            string appDirectory)
+            string appDirectory = null)
         {
             // Load user assemblies from the nested "bin" folder
             FunctionEndpoint.LoadAssemblies(appDirectory);
 
-            var startableEndpoint = EndpointWithExternallyManagedServiceProvider.Create(
-                    configuration.EndpointConfiguration,
+            return EndpointWithExternallyManagedServiceProvider.Create(
```
andreasohlund, line 70: This is a behaviour change since we are creating the endpoint earlier but myself and @danielmarbach believes this is an improvement since configurations will surface immediately rather than on the first invocation.

src/NServiceBus.AzureFunctions.SourceGenerator.Tests/ApprovalFiles/SourceGeneratorApprovals2.Can_enable_transactions.approved.txt
```diff
@@ -10,9 +10,9 @@ class FunctionEndpointTrigger
 {
     readonly FunctionEndpoint endpoint;
 
-    public FunctionEndpointTrigger(FunctionEndpoint endpoint)
+    public FunctionEndpointTrigger(IFunctionEndpoint endpoint)
     {
-        this.endpoint = endpoint;
+        this.endpoint = (FunctionEndpoint) endpoint;
```
andreasohlund, line 15: We need to cast here since the MS bug prevents us from injecting `FunctionEndpoint` and we don't want to break the public API in a minor

src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
```diff
@@ -118,9 +118,9 @@ class FunctionEndpointTrigger
 {{
     readonly FunctionEndpoint endpoint;
 
-    public FunctionEndpointTrigger(FunctionEndpoint endpoint)
+    public FunctionEndpointTrigger(IFunctionEndpoint endpoint)
     {{
-        this.endpoint = endpoint;
+        this.endpoint = (FunctionEndpoint) endpoint;
```
andreasohlund, line 123: We need to cast here since the MS bug prevents us from injecting `FunctionEndpoint` and we don't want to break the public API in a minor

DavidBoike (approved):

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
```diff
@@ -22,8 +22,13 @@
     /// </summary>
     public class FunctionEndpoint : IFunctionEndpoint
```
danielmarbach, line 23: Maybe give them a heads-up about the obsoletion?

https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/blob/master/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/obsoletes-v3.cs#L16

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
danielmarbach, line 31: Maybe same here since we will internalize this type again

src/NServiceBus.AzureFunctions.SourceGenerator/TriggerFunctionGenerator2.cs
danielmarbach: Isn't the real reason here that we don't want to call the public API here because of the nasty reflection/stack frame walking that would then be necessary internally?

andreasohlund: That is true

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
andreasohlund: That is a good idea, not sure what the message should be since there isn't really anything users could do right now due to the reflection stuff that will slow things down if calling the `IFunctionEndpoint` API?

I wonder if it would be better to not add it?

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
andreasohlund: If we obsolete the entire type this shouldn't be needed right?

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
danielmarbach: Yeah you are right unless we would have an extension method that accepts the new parameters to mimic the new API that expects the flag and internally casts things. Then people would already be prepared for the v3

andreasohlund: Given how rare it would be for users to actually call this API manually I'm thinking YAGNI?

andreasohlund:
I've tested this manually by deploying both the v3 and v4 host projects to azure and everything seems to be ok. Also tested the v4 host with both net5 and net6 and that also worked as expected

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionEndpoint.cs
andreasohlund: Api is now obsoleted and I have added an extension method to help users with migration like we discussed

danielmarbach (approved):

PR #401, opened 2022-01-11
[Spike] Less strongly typed public api that is less prone to versioning issues

andreasohlund:
This is based on @bording idea to rely on our sourcegenerator to hide the "ugly" API from our customers. This has the upside of:

- Doesn't have to change when MS changes SDK types
- Doesn't use an interface which means less risk to break

As an aside, this achieves interface segregation where users would continue to use `IFunctionEndpoint` to perform their message operations outside the message handling pipeline just like today without having to see the process method which they shouldn't be using.

The only negative effect is that customers that DO have to use the workaround to call the process methods manually would have less type safety but that is a small price to pay and we should strive to evolve our source generator to cater for their needs anyway to minimize the need to use the workaround.

### DIFF
--- a/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/IFunctionEndpoint.cs
+++ b/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/IFunctionEndpoint.cs
@@ -3,8 +3,6 @@
     using System;
     using System.Threading;
     using System.Threading.Tasks;
-    using Azure.Messaging.ServiceBus;
-    using Microsoft.Azure.WebJobs.ServiceBus;
     using Microsoft.Extensions.Logging;
     using ExecutionContext = Microsoft.Azure.WebJobs.ExecutionContext;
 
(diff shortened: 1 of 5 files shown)


danielmarbach:
While I think it is OK to go down that path, I feel we are trying to do something just for keeping the holy grail of SemVer
Since it is a downstream, I really don't think it is problematic to have more majors
for me, that is the lesser evil than going down the route of almost dynamic programming :smile:
A similar design would be to have a context instead of an object that has typed information as well as an extension bag with less typed information that would allow floating in information from the generator part to the core

timbussmann:
> Doesn't have to change when MS changes SDK types

not really sure whether I fully understand this. The code still casts all the passed `objects` to back to specific types, so that would still fail at runtime if the SDK types change?

SzymonPobiega:
I am with @danielmarbach , I think that the added complexity has more cost than the benefits of reducing the risk of need for breaking changes. The next person who comes to this repo would have hard time understanding why all these casting.

DavidBoike:
It's about more than just avoiding breaking changes though. The value-add we bring here is making a specific hosting+messaging infrastructure easier for the end-user. So it's not about Particular wanting to avoid having to cut a bunch of major releases because of breaking changes. It's about making it easier for the end user to upgrade to the latest SDKs (a moving target) by just updating to the next minor version of our package and not having to change any of their own code, absorbing some of the complexity for them so they can focus on what really matters.

> The code still casts all the passed objects to back to specific types, so that would still fail at runtime if the SDK types change?

Yes, we'd release a new minor that would use the new SDK version and the customer wouldn't need to change anything. Win!

> The next person who comes to this repo would have hard time understanding why all these casting.

That's what code comments are for.

danielmarbach:
What SDK upgrade are we talking about? SendsWithAtomicReceive was the last missing part that was only late introduced into the service bus SDK. Now it is near feature complete from a public API perspective. It is hard to predict the future but with the third incarnation of the SDKs (or is it the fourth, counting ServiceBus for Windows Services?) I doubt there will be many more groundbreaking changes to the SDK. It is probably much more likely at this stage that functions SDK will evolve, and in such cases we might need to adjust anyway. I'm just not seeing yet how this change makes things easier for customers to upgrade, but I might be missing something. 

If we go down the route, my preference would be on a context like bag design, but again that is just pure taste from my side.

PR #404, opened 2022-01-19
Obsolete APIs for next major

aleksandr-samila:
Restored obsolete APIs `ServiceBusTriggeredEndpointConfiguration` constructors which are being removed and a few more

### DIFF
--- a/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/obsoletes-v4.cs
+++ b/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/obsoletes-v4.cs
@@ -14,45 +14,30 @@ namespace NServiceBus
                     RemoveInVersion = "5")]
         public string ServiceBusConnectionString { get; set; }
 
-        /// <summary>
-        /// Creates a serverless NServiceBus endpoint.
-        /// </summary>
         [ObsoleteEx(
             TreatAsErrorFromVersion = "4",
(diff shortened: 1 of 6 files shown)


src/NServiceBus.AzureFunctions.InProcess.ServiceBus/obsoletes-v4.cs
```diff
@@ -2,11 +2,75 @@
...
     public partial class ServiceBusTriggeredEndpointConfiguration
     {
         [ObsoleteEx(ReplacementTypeOrMember = "UseNServiceBus(ENDPOINTNAME, CONNECTIONSTRING)",
                     TreatAsErrorFromVersion = "4",
                     RemoveInVersion = "5")]
         public string ServiceBusConnectionString { get; set; }
+
+        /// <summary>
+        /// Creates a serverless NServiceBus endpoint.
+        /// </summary>
```
mikeminutillo, line 19: I think in these obsolete-only files we don't need XML docs. We can disable the check with `#pragma warning disable 1591` at the top of the file (with `#pragma warning restore 1591` at the bottom of the file to turn it back on for the rest of the solution. See the [core obsoletes](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/obsoletes-v8.cs) for more info

mikeminutillo (approved):

PR #439, opened 2022-03-15
Update to core v8 beta 1

timbussmann:


src/NServiceBus.AzureFunctions.InProcess.ServiceBus/Serverless/TransportWrapper/PipelineInvoker.cs
```diff
@@ -28,6 +28,9 @@ public PipelineInvoker(IMessageReceiver baseTransportReceiver)
 
         public Task StartReceive(CancellationToken cancellationToken) => Task.CompletedTask;
 
+        // Noop because the rate at which Azure Functions pushes messages to the pipeline can't be controlled.
+        public Task ChangeConcurrency(PushRuntimeSettings limitations, CancellationToken cancellationToken = new CancellationToken()) => Task.CompletedTask;
```
timbussmann, line 32: not sure whether this should instead throw an exception in case the user enables this behavior but I think that should be done outside of the scope of this PR.

HEskandari: @timbussmann Should we raise an issue for it?

mikeminutillo (approved):

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/Serverless/TransportWrapper/PipelineInvoker.cs
timbussmann: I was thinking a bit more about this and I think `CompletedTask` is the right choice. Throwing an exception would only happen once the endpoint recoverability would try to change the concurrency, which would be potentially late at runtime and then the exception would only cause further chaos, which should be avoided

It should be considered to throw at startup or prevent the configuration of this recoverability feature, which with Azure Functions is buried within deeper configuration levels anyway. I don't think there is something worthwhile doing here, there are other endpoint configuration options that shouldn't be used when running on azure functions that don't strictly throw exceptions

PR #527, opened 2022-09-20
Making the referencers of the library treat CS8032 warning as error

cquirosj:
Making the referencers of the library treat CS8032 warning as an error. CS8032 warning happens when a source code generator fails to load


kbaley (approved):

kbaley:
This will be included when 4.1.0 is released

PR #571, opened 2022-12-01
Host throws assembly scanning errors on startup

andreasohlund:
This fixes the bug but we should make sure we can repro with a unit test before merging this.

https://microsoft.github.io/AzureTipsAndTricks/blog/tip196.html and https://stackoverflow.com/questions/59743560/how-to-unit-test-azure-functions-in-visual-studio show how to use func.exe but there might be other better ways

### DIFF
--- a/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/InProcessFunctionEndpoint.cs
+++ b/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/InProcessFunctionEndpoint.cs
@@ -239,7 +239,9 @@
             "Grpc.Net.ClientFactory.dll",
             "Azure.Identity.dll",
             "Microsoft.Extensions.Azure.dll",
-            "NServiceBus.Extensions.DependencyInjection.dll"
+            "NServiceBus.Extensions.DependencyInjection.dll",
+            "Microsoft.Identity.Client.dll",
+            "Microsoft.Identity.Client.Extensions.Msal.dll"
         };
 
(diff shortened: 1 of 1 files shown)


src/NServiceBus.AzureFunctions.InProcess.ServiceBus/InProcessFunctionEndpoint.cs
```diff
@@ -239,7 +239,9 @@ static async Task SafeCompleteMessageAsync(ServiceBusMessageActions messageActio
             "Grpc.Net.ClientFactory.dll",
             "Azure.Identity.dll",
             "Microsoft.Extensions.Azure.dll",
-            "NServiceBus.Extensions.DependencyInjection.dll"
+            "NServiceBus.Extensions.DependencyInjection.dll",
```
andreasohlund, line 242: This feels like a never-ending story of patches as MS adds more references to the azure host but not sure how to make this better

DavidBoike: Really? That's all? Would this work in InternalAutomation if I added these excludes in the NSB config?

For CI you can [install the functions core tools with automated tools](https://github.com/Azure/azure-functions-core-tools#installing)

andreasohlund: Yeah, that should work

Thanks for the link👍

SzymonPobiega (approved):

andreasohlund:
@SzymonPobiega @WojcikMike ready for final review

SzymonPobiega (approved):

PR #578, opened 2022-12-07
Propagate the atomic sends flag to config

andreasohlund:


andreasohlund:
We'll go with a different approach

PR #596, opened 2022-12-15
Add an end-to-end test using func.exe to make sure assembly scanning works

SzymonPobiega:
The test app has been switched to use .NET 6 as a target because .NET 7 is not supported in the isolated model ([only LTS versions are going to be supported, according to MS](https://github.com/Azure/azure-functions-host/issues/8909#issuecomment-1334640564))


andreasohlund (approved):

andreasohlund:
@SzymonPobiega all green, please do a final review

PR #614, opened 2023-01-16
Defer serverless set up until after customization applied

mikeminutillo:
This is the in process equivalent of https://github.com/Particular/NServiceBus.AzureFunctions.Worker.ServiceBus/pull/255

Defers setting up the serverless infrastructure until after the user customization has been applied. After this change, it doesn't matter if `UseTransport(..)` is called, as that configuration is overwritten just before the endpoint is started.

### DIFF
--- a/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionsHostBuilderExtensions.cs
+++ b/src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionsHostBuilderExtensions.cs
@@ -81,6 +81,7 @@
         static void RegisterEndpointFactory(IFunctionsHostBuilder functionsHostBuilder,
             ServiceBusTriggeredEndpointConfiguration serviceBusTriggeredEndpointConfiguration)
         {
+            var serverless = serviceBusTriggeredEndpointConfiguration.MakeServerless();
             // When using functions, assemblies are moved to a 'bin' folder within FunctionsHostBuilderContext.ApplicationRootPath.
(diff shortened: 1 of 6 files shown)


src/NServiceBus.AzureFunctions.InProcess.ServiceBus/FunctionsHostBuilderExtensions.cs
```diff
@@ -81,6 +81,7 @@ public static partial class FunctionsHostBuilderExtensions
         static void RegisterEndpointFactory(IFunctionsHostBuilder functionsHostBuilder,
             ServiceBusTriggeredEndpointConfiguration serviceBusTriggeredEndpointConfiguration)
         {
+            var serverless = serviceBusTriggeredEndpointConfiguration.MakeServerless();
```
SzymonPobiega, line 84: Isn't that too early? [In the Worker](https://github.com/Particular/NServiceBus.AzureFunctions.Worker.ServiceBus/pull/255/files#diff-06330f09da8779aa1583b4ed457a9884255dc9b56ee7b88be1db23dd61cba02eR141) it is invoked in a factory method registered in the DI and here it is executed immediately in the `UseNserviceBus`

mikeminutillo: As long as it is invoked after the user configuration is applied and before the startable endpoint is created, it will work.

SzymonPobiega (approved):

PR #641, opened 2023-03-21
Properly test atomic sends with receive

andreasohlund:


src/IntegrationTests.HostV4/host.json
```diff
@@ -1,3 +1,8 @@
 {
-    "version": "2.0"
+  "version": "2.0",
+  "extensions": {
+    "ServiceBus": {
+      "EnableCrossEntityTransactions": true
```
andreasohlund, line 5: @danielmarbach this is the fix

src/IntegrationTests.HostV4/IntegrationTests.HostV4.csproj
```diff
@@ -14,6 +14,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="4.1.3" />
+    <PackageReference Include="NServiceBus" Version="8.0.3" />
+    <PackageReference Include="NServiceBus.Transport.AzureServiceBus" Version="3.2.0" />
```
andreasohlund, line 18: We add explicit refs for all deps where the main package has a range right?

danielmarbach (approved):

Issue #651, opened 2023-04-03
Some advanced NSB configuration APIs aren't applicable to serverless environments


kbaley:
### Describe the suggested improvement

For the AzureFunctions.Worker.ServiceBus package, much of the NServiceBus configuration API has been moved behind the `AdvancedConfiguration` API to keep users focused on the most common ones. Some of these APIs aren't applicable to a serverless environment but there's nothing stopping someone from using them, sometimes with unpredictable results. Some examples:

- UseTransport
- PurgeOnStartup
- DefineCriticalErrorAction
- SetDiagnosticsPath
- LimitMessageProcessingConcurrencyTo
- MakeInstanceUniquelyAddressable
- OverrideLocalAddress
- RouteReplyToThisInstance
- RouteToThisInstance

Plus some Azure Service Bus transport properties:
- MaxAutoLockRenewalDuration
- PrefetchCount
- PrefetchMultiplier
- TimeToWaitBeforeTriggeringCircuitBreaker

We should implement a way to ensure these APIs aren't used in production. 

### Additional Context

A spike for a Roslyn analyzer that throws an error when certain APIs are used [can be found here](https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/pull/640).

- https://github.com/Particular/NServiceBus.AwsLambda.Sqs/issues/253
- https://github.com/Particular/NServiceBus.AzureFunctions.Worker.ServiceBus/issues/302

PR #717, opened 2023-09-08
Add Rosylyn analyzer for irrelevant APIs

jpalac:
- Rebase of https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/pull/640
- Fixes https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/issues/651


src/NServiceBus.AzureFunctions.Analyzer.Tests/AzureFunctionsTransportAnalyzerTests.cs
```diff
@@ -0,0 +1,40 @@
...
+        [|transportConfig.{configName}|] = {configValue};
+    }}
+
+    void Extension(TransportExtensions<AzureServiceBusTransport> transportExtension)
+    {{
+        [|transportExtension.{configName}({configValue})|];
+    }}
+}}";
+
+            return Assert(diagnosticId, source);
```
timbussmann, line 37: would it be worthwhile to have a test case that this analyzer doesn't trigger on other classes with properties that have the same name? This comment also applies to all other tested APIs I think.

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsConfigurationAnalyzer.cs
```diff
@@ -0,0 +1,176 @@
...
+            }
+
+            var memberAccessSymbol = context.SemanticModel.GetSymbolInfo(memberAccessExpression, context.CancellationToken);
+
+            if (!(memberAccessSymbol.Symbol is IMethodSymbol methodSymbol))
+            {
+                return;
+            }
+
+            if (methodSymbol.ReceiverType.ToString() == "NServiceBus.EndpointConfiguration")
```
timbussmann, line 130: would that also work with extension methods?

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
```diff
@@ -0,0 +1,162 @@
...
+        public const string TimeToWaitBeforeTriggeringCircuitBreakerNotAllowedId = "NSBFUNC016";
+
+        public const string EntityMaximumSizeNotAllowedId = "NSBFUNC017";
+
+        const string DiagnosticCategory = "NServiceBus.AzureFunctions";
+
+        internal static readonly DiagnosticDescriptor PurgeOnStartupNotAllowed = new DiagnosticDescriptor(
+             id: PurgeOnStartupNotAllowedId,
+             title: "PurgeOnStartup is not supported in Azure Functions",
+             messageFormat: "Azure Functions endpoints are started when the first message arrives. PurgeOnStartup may purge whenever a new instance is started.",
```
timbussmann, line 30: I think the idea here is correct, however there are two issues which should be considered:

1. ASB doesn't support PurgeOnStartup at all as it seems https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/master/src/Transport/Receiving/MessagePump.cs#L56
2. Purging would happen when the receiver (pump) is started, with Azure Functions, the call to start the pump is actually swallowed and the pump isn't started at all.

so it might be easier to just remain on the "not supported by ASB" message for now (this exception would never show up in Azure Functions at the moment but it would on regular ASB usages).

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
```diff
@@ -0,0 +1,162 @@
...
+             category: DiagnosticCategory,
+             defaultSeverity: DiagnosticSeverity.Error,
+             isEnabledByDefault: true
+            );
+
+        internal static readonly DiagnosticDescriptor LimitMessageProcessingToNotAllowed = new DiagnosticDescriptor(
+             id: LimitMessageProcessingToNotAllowedId,
+             title: "LimitMessageProcessing is not supported in Azure Functions",
+             messageFormat: "Azure Functions endpoints do not control the message receiver and cannot limit message processing concurrency.",
```
timbussmann, line 39: https://learn.microsoft.com/en-us/azure/azure-functions/functions-concurrency#static-concurrency states that the ASB trigger offers concurrency settings on the trigger. I don't see the mentioned settings being documented in the [trigger documentation](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-service-bus-trigger?tabs=python-v2%2Cin-process%2Cnodejs-v4%2Cextensionv5&pivots=programming-language-csharp) but that might be worth checking?

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
```diff
@@ -0,0 +1,162 @@
...
+             category: DiagnosticCategory,
+             defaultSeverity: DiagnosticSeverity.Error,
+             isEnabledByDefault: true
+            );
+
+        internal static readonly DiagnosticDescriptor DefineCriticalErrorActionNotAllowed = new DiagnosticDescriptor(
+             id: DefineCriticalErrorActionNotAllowedId,
+             title: "DefineCriticalErrorAction is not supported in Azure Functions",
+             messageFormat: "Azure Functions endpoints do not control the application lifecycle and should not define behavior in the case of critical errors.",
```
timbussmann, line 48: IIRC another reason was also that critical errors are primarily raised by the message pump, which is not active here.

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
```diff
@@ -0,0 +1,162 @@
...
+             messageFormat: "Azure Functions endpoints should not write diagnostics to the local file system. Use CustomDiagnosticsWriter to write diagnostics to another location.",
+             category: DiagnosticCategory,
+             defaultSeverity: DiagnosticSeverity.Error,
+             isEnabledByDefault: true
+            );
+
+        internal static readonly DiagnosticDescriptor MakeInstanceUniquelyAddressableNotAllowed = new DiagnosticDescriptor(
+             id: MakeInstanceUniquelyAddressableNotAllowedId,
+             title: "MakeInstanceUniquelyAddressable is not supported in Azure Functions",
+             messageFormat: "Azure Functions endpoints have unpredictable lifecycles and should not be uniquely addressable.",
```
timbussmann, line 66: would it help to point out the risk of message loss due to that? A bit on the fence tbh, I think the current message might be a good start.

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
```diff
@@ -0,0 +1,162 @@
...
+             messageFormat: "Azure Functions endpoints do not control the message receiver and cannot decide the local address.",
+             category: DiagnosticCategory,
+             defaultSeverity: DiagnosticSeverity.Error,
+             isEnabledByDefault: true
+            );
+
+        internal static readonly DiagnosticDescriptor RouteReplyToThisInstanceNotAllowed = new DiagnosticDescriptor(
+             id: RouteReplyToThisInstanceNotAllowedId,
+             title: "RouteReplyToThisInstance is not supported in Azure Functions",
+             messageFormat: "Azure Functions endpoints do not control the message receiver and cannot configure specific instance routing.",
```
timbussmann, line 93: it's not really about controlling the receiver but that for "RouteReplyToThisInstance" the endpoints needs to be uniquely addressable which is already prevented. Maybe state something like `Azure Functions instances cannot be directly addressed as they have a highly volatile lifetime. Use 'RouteToThisEndpoint' instead`

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
```diff
@@ -0,0 +1,162 @@
...
+             category: DiagnosticCategory,
+             defaultSeverity: DiagnosticSeverity.Error,
+             isEnabledByDefault: true
+            );
+
+        internal static readonly DiagnosticDescriptor RouteToThisInstanceNotAllowed = new DiagnosticDescriptor(
+             id: RouteToThisInstanceNotAllowedId,
+             title: "RouteToThisInstance is not supported in Azure Functions",
+             messageFormat: "Azure Functions endpoints do not control the message receiver and cannot configure specific instance routing.",
```
timbussmann, line 102: same as comment above

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
```diff
@@ -0,0 +1,162 @@
...
+             category: DiagnosticCategory,
+             defaultSeverity: DiagnosticSeverity.Error,
+             isEnabledByDefault: true
+            );
+
+        internal static readonly DiagnosticDescriptor RouteReplyToAnyInstanceNotAllowed = new DiagnosticDescriptor(
```
timbussmann, line 108: I think this API should actually be allowed? According to https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.AcceptanceTests/Routing/When_routing_reply_to_any_instance.cs#L13, this API should route the reply to the shared queue which is stable.

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
```diff
@@ -0,0 +1,162 @@
...
+             messageFormat: "Azure Functions endpoints do not control the message receiver and by default route the replies to any instance.",
+             category: DiagnosticCategory,
+             defaultSeverity: DiagnosticSeverity.Warning,
+             isEnabledByDefault: true
+            );
+
+        internal static readonly DiagnosticDescriptor MaxAutoLockRenewalDurationNotAllowed = new DiagnosticDescriptor(
+             id: MaxAutoLockRenewalDurationNotAllowedId,
+             title: "MaxAutoLockRenewalDuration is not supported in Azure Functions",
+             messageFormat: "Azure Functions endpoints do not control the message receiver and cannot decide the lock renewal duration.",
```
timbussmann, line 120: should we refer to the trigger specific config option instead? https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-service-bus-trigger?tabs=python-v2%2Cin-process%2Cnodejs-v4%2Cextensionv5&pivots=programming-language-csharp#peeklock-behavior

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
```diff
@@ -0,0 +1,162 @@
...
+             messageFormat: "Azure Functions endpoints do not control the message receiver and cannot decide the prefetch count.",
+             category: DiagnosticCategory,
+             defaultSeverity: DiagnosticSeverity.Error,
+             isEnabledByDefault: true
+            );
+
+        internal static readonly DiagnosticDescriptor PrefetchMultiplierNotAllowed = new DiagnosticDescriptor(
+             id: PrefetchMultiplierNotAllowedId,
+             title: "PrefetchMultiplier is not supported in Azure Functions",
+             messageFormat: "Azure Functions endpoints do not control the message receiver and cannot decide the prefetch multiplier.",
```
timbussmann, line 138: rather than saying that Azure Function endpoints do not control stuff, which might be confusing to users, should we instead say something like `Message prefetching is controlled by the Azure Service Bus trigger and cannot be configured via the NServiceBus transport configuration API when using Azure Functions`?

timbussmann (commented):
Looks solid to me. I added a few thoughts about the diagnostic messages (sorry for not making better suggestions right away).

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
```diff
@@ -0,0 +1,162 @@
...
+             messageFormat: "The package configures Azure Service Bus transport by default. Use ServiceBusTriggeredEndpointConfiguration.Transport to access the transport configuration.",
+             category: DiagnosticCategory,
+             defaultSeverity: DiagnosticSeverity.Warning,
+             isEnabledByDefault: true
+            );
+
+        internal static readonly DiagnosticDescriptor OverrideLocalAddressNotAllowed = new DiagnosticDescriptor(
+             id: OverrideLocalAddressNotAllowedId,
+             title: "OverrideLocalAddress is not supported in Azure Functions",
+             messageFormat: "Azure Functions endpoints do not control the message receiver and cannot decide the local address.",
```
danielmarbach, line 84: Wouldn't the better explanation be that the address is determined by the endpoint name? https://docs.particular.net/nservicebus/hosting/azure-functions-service-bus/#basic-usage-azure-function-queue-trigger-for-nservicebus

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
danielmarbach: Agreed

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
danielmarbach: Especially because our docs say

> Concurrency-related settings are controlled via the Azure Function host.json configuration file. See [Concurrency in Azure Functions](https://docs.microsoft.com/en-us/azure/azure-functions/functions-concurrency#service-bus) for details.

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
```diff
@@ -0,0 +1,162 @@
+﻿namespace NServiceBus.AzureFunctions.Analyzer
+{
+    using Microsoft.CodeAnalysis;
+
+    public static class AzureFunctionsDiagnostics
```
danielmarbach, line 5: Given that consistency is automatically set by the generator should we also prevent the transaction mode to be explicitly set?

https://docs.particular.net/nservicebus/hosting/azure-functions-service-bus/#message-consistency

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
```diff
@@ -0,0 +1,162 @@
...
+             title: "TimeToWaitBeforeTriggeringCircuitBreaker is not supported in Azure Functions",
+             messageFormat: "Azure Functions endpoints do not control the message receiver and cannot access circuit breaker settings.",
+             category: DiagnosticCategory,
+             defaultSeverity: DiagnosticSeverity.Error,
+             isEnabledByDefault: true
+            );
+
+        internal static readonly DiagnosticDescriptor EntityMaximumSizeNotAllowed = new DiagnosticDescriptor(
+             id: EntityMaximumSizeNotAllowedId,
+             title: "EntityMaximumSize is not supported in Azure Functions",
```
danielmarbach, line 155: As mentioned above this setting is for queues and topics unfortunately similar to enable partitioning

https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/883c4bd2c7becbaa8fa51a7a179e0f6178ffa73c/src/Transport/AzureServiceBusTransport.cs#L154

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
danielmarbach: I guess UseWebSockets is also not allowed since the connection is established by the host

https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/883c4bd2c7becbaa8fa51a7a179e0f6178ffa73c/src/Transport/AzureServiceBusTransport.cs#L296

that would also prevent the web proxy I guess? https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/blob/883c4bd2c7becbaa8fa51a7a179e0f6178ffa73c/src/Transport/AzureServiceBusTransport.cs#L315

but I didn't have time to verify all that with all the connection mechanics that are supported (or not). Might be worth checking

src/NServiceBus.AzureFunctions.Analyzer.Tests/AzureFunctionsConfigurationAnalyzerTests.cs
```diff
@@ -0,0 +1,37 @@
...
+    void Bar(ServiceBusTriggeredEndpointConfiguration endpointConfig)
+    {{
+        [|endpointConfig.AdvancedConfiguration.{configuration}|];
+
+        var advancedConfig = endpointConfig.AdvancedConfiguration;
+        [|advancedConfig.{configuration}|];
+    }}
+}}";
+
+            return Assert(diagnosticId, source);
```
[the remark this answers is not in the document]
jpalac: Having the 2 sets of "[|" means that 2 warnings are expected. If only one was given then the test would fail.

src/NServiceBus.AzureFunctions.Analyzer.Tests/AzureFunctionsTransportAnalyzerTests.cs
jpalac: We have added some tests for this.

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
jpalac: removed

src/NServiceBus.AzureFunctions.Analyzer.Tests/AzureFunctionsConfigurationAnalyzerTests.cs
```diff
@@ -0,0 +1,37 @@
+namespace NServiceBus.AzureFunctions.Analyzer.Tests
+{
+    using System.Threading.Tasks;
+    using NUnit.Framework;
+    using static AzureFunctionsDiagnostics;
+
+    [TestFixture]
+    public class AzureFunctionsConfigurationAnalyzerTests : AnalyzerTestFixture<AzureFunctionsConfigurationAnalyzer>
```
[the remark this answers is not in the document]
jpalac: We have renamed the tests

src/NServiceBus.AzureFunctions.Analyzer.Tests/AzureFunctionsConfigurationAnalyzerTests.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        [TestCase("DefineCriticalErrorAction((errorContext, cancellationToken) => Task.CompletedTask)", DefineCriticalErrorActionNotAllowedId)]
+        [TestCase("LimitMessageProcessingConcurrencyTo(5)", LimitMessageProcessingToNotAllowedId)]
+        [TestCase("MakeInstanceUniquelyAddressable(null)", MakeInstanceUniquelyAddressableNotAllowedId)]
+        [TestCase("OverrideLocalAddress(null)", OverrideLocalAddressNotAllowedId)]
+        [TestCase("PurgeOnStartup(true)", PurgeOnStartupNotAllowedId)]
```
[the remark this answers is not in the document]
jpalac: It does not - calling the method in general should not be allowed.

src/NServiceBus.AzureFunctions.Analyzer.Tests/AzureFunctionsSendReplyOptionsAnalyzerTests.cs
```diff
@@ -0,0 +1,30 @@
+namespace NServiceBus.AzureFunctions.Analyzer.Tests
+{
+    using System.Threading.Tasks;
+    using NUnit.Framework;
+    using static AzureFunctionsDiagnostics;
+
+    [TestFixture]
+    public class AzureFunctionsSendReplyOptionsAnalyzerTests : AnalyzerTestFixture<AzureFunctionsConfigurationAnalyzer>
```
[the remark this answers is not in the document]
jpalac: We have renamed the file

src/NServiceBus.AzureFunctions.Analyzer.Tests/AzureFunctionsTransportAnalyzerTests.cs
```diff
@@ -0,0 +1,40 @@
+namespace NServiceBus.AzureFunctions.Analyzer.Tests
+{
+    using System.Threading.Tasks;
+    using NUnit.Framework;
+    using static AzureFunctionsDiagnostics;
+
+    [TestFixture]
+    public class AzureFunctionsTransportAnalyzerTests : AnalyzerTestFixture<AzureFunctionsConfigurationAnalyzer>
+    {
+        [TestCase("EntityMaximumSize", "5", EntityMaximumSizeNotAllowedId)]
```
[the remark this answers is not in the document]
jpalac: Originally we thought that only EntityMaximumSize  was used during queue creation - we have now added EnablePartitioning as well

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
jpalac: We have reworded this.

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
jpalac: We have added an EnablePartitioning diagnostic

src/NServiceBus.AzureFunctions.Analyzer.Tests/AnalyzerTestFixture.cs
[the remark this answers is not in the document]
jpalac: correct

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsConfigurationAnalyzer.cs
jpalac: We have tested transport extension methods.
What we have not done a diagnostic for is direct use of the static methods ie ConfigurePurging.PurgeOnStartup() but we don't think it's worth it as no one should be using that approach. We can revisit if that proves to be wrong.

src/NServiceBus.AzureFunctions.Analyzer/AzureFunctionsDiagnostics.cs
jpalac: We have added TransportTransactionMode to the diagnostics.

We are not sure about the websockets and webproxy so thought we may leave them out and revisit if they prove to be a problem as well.

jpalac:
Thank you @timbussmann and @danielmarbach for taking the time to look over this PR.

mikeminutillo (approved):

PR #866, opened 2024-11-27
Update to NServiceBus 9

bording:
This PR updates the repo to a new major version that works with NServiceBus 9.

It also makes changes to combine analyzer and source generator projects, similar to the changes made to the Worker package: https://github.com/Particular/NServiceBus.AzureFunctions.Worker.ServiceBus/pull/391


andreasohlund:
Looking at the logs the only thing that looks odd is https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/pull/866/checks#step:12:135

but then later we can see that the [HTTP request to initiate the test works](https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/pull/866/checks#step:12:322) but then nothing more happens. Is it possible to check if there are some messages in the servicebus namespace used by the test?

andreasohlund:
I can repro this locally. The httptrigger is invoked and I can see a message in the input queue but it doesn't get processed so it seems like the binding isn't triggering a receive of the message.
I'll keep digging

andreasohlund:
Seems like things start up but the generated servicebus trigger is not detected.

andreasohlund:
On master it does get picked up:

```
Functions:

        InProcessHttpSenderV4: [GET,POST] http://localhost:7071/api/InProcessHttpSenderV4

        NServiceBusFunctionEndpointTrigger-InProcess-HostV4: serviceBusTrigger

```

but not on this branch

```
Functions:

        InProcessHttpSenderV4: [GET,POST] http://localhost:7071/api/InProcessHttpSenderV4

```

src/IntegrationTests.HostV4/IntegrationTests.HostV4.csproj
```diff
@@ -1,15 +1,14 @@
...
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
     <AzureFunctionsVersion>v4</AzureFunctionsVersion>
     <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
     <CompilerGeneratedFilesOutputPath>$(BaseIntermediateOutputPath)\GeneratedFiles</CompilerGeneratedFilesOutputPath>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\NServiceBus.AzureFunctions.InProcess.ServiceBus\NServiceBus.AzureFunctions.InProcess.ServiceBus.csproj" />
-    <ProjectReference Include="..\NServiceBus.AzureFunctions.SourceGenerator\NServiceBus.AzureFunctions.SourceGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
```
andreasohlund, line 12: After debugging this further is seems like the code gen isn't kicking in and the function is not generated, could this be the culprit? cc @danielmarbach

danielmarbach: Judging by the PR the project was renamed. As far as I recall it need to be referenced because they are not automatically included when adding a project reference (at least they weren't in the past) but it is a while ago since I looked at the project

bording: Could we be hitting some of the same problems with their generator that that we saw in Worker that caused us to need to set [this stuff](https://github.com/Particular/NServiceBus.AzureFunctions.Worker.ServiceBus/blob/master/src/NServiceBus.AzureFunctions.Worker.ServiceBus/NServiceBus.AzureFunctions.Worker.ServiceBus.props)?

I'll dig into that a bit now that I have some idea of why it might not be working.

danielmarbach: Could very well be yes

bording: Sigh, that project reference shouldn't have been removed. Visual Studio must have done that behind my back and I didn't catch it.

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/NServiceBus.AzureFunctions.InProcess.ServiceBus.csproj
```diff
@@ -1,43 +1,36 @@
...
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.Functions.Extensions" Version="1.1.0" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.ServiceBus" Version="5.16.4" />
-    <PackageReference Include="NServiceBus.Newtonsoft.Json" Version="3.0.0" />
-    <PackageReference Include="NServiceBus" Version="8.2.4" />
-    <PackageReference Include="NServiceBus.Transport.AzureServiceBus" Version="3.2.5" />
-    <PackageReference Include="Obsolete.Fody" Version="5.3.0" PrivateAssets="All" />
-    <PackageReference Include="Particular.Packaging" Version="4.2.0" PrivateAssets="All" />
+    <PackageReference Include="NServiceBus" Version="9.2.3" />
+    <PackageReference Include="NServiceBus.Newtonsoft.Json" Version="4.0.0" />
```
andreasohlund, line 17: Should we switch to system.json assuming this will go in a new major?

andreasohlund (approved):

src/NServiceBus.AzureFunctions.InProcess.ServiceBus/NServiceBus.AzureFunctions.InProcess.ServiceBus.csproj
bording: I wasn't sure what it was used for and if there would be wire compat concerns if I changed it. I'm all for switching it if we can.

andreasohlund: To switch we need to stop setting json.net as the default and require the user explicitly set the serializer. I wonder if that is worth the hazzle given that this package is considered legacy?

bording: After discussion, we aren't going to change this.

andreasohlund:
The emulator fails with

>   [2025-02-14T06:48:30.955Z] Microsoft.Azure.WebJobs.Host: Error indexing method 'NServiceBusFunctionEndpointTrigger-InProcess-HostV4'. System.Private.CoreLib: Could not load file or assembly 'System.Memory.Data, Version=6.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.

So it seems like something is still referencing net6?

andreasohlund:
Seems to be https://github.com/Azure/azure-functions-host/issues/10575

I have applied the mentioned workaround

PR #893, opened 2025-05-21
Prepare for NServiceBus 10

tamararivera:
Prepare for NServiceBus 10


bording:
.NET 8 is the last version that the in-process model supports, and support ends for the in-process itself ends in [November 2026](https://azure.microsoft.com/en-us/updates?id=retirement-support-for-the-inprocess-model-for-net-apps-in-azure-functions-ends-10-november-2026), so it won't be possible to update it to .NET 10 / NServiceBus 10.

PR #1020, opened 2026-06-15
Guide towards the new package

andreasohlund:


README.md
```diff
@@ -2,11 +2,13 @@
 > **NServiceBus.AzureFunctions.InProcess.ServiceBus has been officially sunsetted.**  
 > Microsoft announced that .NET 8 will be [the last release supporting the in-process hosting model](https://techcommunity.microsoft.com/t5/apps-on-azure-blog/net-on-azure-functions-august-2023-roadmap-update/ba-p/3910098).
 > We will continue to provide support and address critical fixes during the sunset period, but no new features will be added.  
-> New projects should use the [isolated worker model](https://docs.particular.net/nservicebus/hosting/azure-functions-service-bus/) instead. Follow [these instructions](https://docs.particular.net/nservicebus/upgrades/azure-functions-service-bus-in-process-isolated-worker) to migrate.
+> New projects should use the [isolated worker model](https://docs.particular.net/nservicebus/hosting/azure/functions/) instead. To migrate:
```
danielmarbach: Could we maybe make this a bit clearer why we recommend them to first upgrade to basically a deprecated project? I think it makes sense from our perspective, but just reading the readme might not be clear. We could also say the recommended procedure is to first upgrade to this one, follow the upgrade guide to the other one because that is the shortest path to go through all the necessary migration paths. Or we can say alternatively you can directly go to the other one, but there is no migration guidance that covers you on that path

andreasohlund: I've fleshed this out a bit now

README.md
```diff
@@ -2,14 +2,23 @@
...
 It is part of the [Particular Service Platform](https://particular.net/service-platform), which includes [NServiceBus](https://particular.net/nservicebus) and tools to build, monitor, and debug distributed systems.
 
+## Migration
+
+To prepare for migration, please review:
+
+- [Migration instructions from InProcess to the Isolated model](https://docs.particular.net/nservicebus/upgrades/azure-functions-service-bus-in-process-isolated-worker).
+- [Migration instructions from the old isolated package to the new multi-endpoint isolated package](https://docs.particular.net/nservicebus/upgrades/azure-functions-service-bus-isolated-to-azure-functions).
+
+Decide whether going straight to the new package is the best course of action, or whether going straight to the latest multi-endpoint package is acceptable from a migration-risk perspective.
```
danielmarbach:
```suggestion
Review both migration guides and decide whether the safer approach is to first migrate from InProcess to the isolated package and then upgrade to the latest multi-endpoint package, or whether a direct migration from InProcess to the latest multi-endpoint package is acceptable.
```

danielmarbach (approved):

